PR #3, opened 2020-03-20
add imagePullSecret to csv

StevenTobin:
The deploymentconfig and subsequent pods created by the csv by default reference an imagePullSecret which will not contain the correct credentials.

We must add an imagePullSecret var to ensure the pods created are able to pull the images.


scripts/delorean/process-csv-images
```diff
@@ -69,6 +70,17 @@ process_csv_images() {
...
 }
 
+update_image_pull_secret() {
+  current_csv=$(get_current_csv_file $PRODUCT)
+  case $PRODUCT in
+  "amq-online")
+    yq w -i ${current_csv} spec.install.spec.deployments[0].spec.template.spec.imagePullSecrets[0].name ${IMAGE_PULL_SECRET}
+    ;;
+  *)
+    echo "skipping pull secret"
```
mikenairn: @StevenTobin Will this need to be done for every product?

StevenTobin: We'll only know that once we get the other branches tested.

PR #9, opened 2020-03-26
Updates around the generation of the image mirror mapping files to avoid conflicts

mikenairn:
* Remove image_mirror_mapping file before re-generating it, but only if the CSV has no references to delorean images.
* Remove check for version equality when extracting manifests from image, we need to update even if the version is the same to ensure we get any updates for pre release versions (operand sha updates etc..).
*  Add image-whitelist, during image processing if the target image exists in this file it will be left as is in the CSV.

### DIFF
--- a/scripts/delorean/image-whitelist
+++ b/scripts/delorean/image-whitelist
@@ -1,0 +1,1 @@
+registry.redhat.io/openshift4/ose-oauth-proxy:4.2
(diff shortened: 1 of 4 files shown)


scripts/delorean/image-whitelist
```diff
@@ -0,0 +1 @@
+registry.redhat.io/openshift4/ose-oauth-proxy:4.2
```
mikenairn: Any images listed in here do not need to be synced as they exist in the correct public repos

PR #13, opened 2020-04-07
DEL-224 Add release osd-addon command to automatically updating the managed-tenants repo

b1zzu:
**What**

- Added the _osd-addon-release_ to automatically update the managed-tenants repo on release

**Steps to verify it**

1. Fork the manage-tenants repo

   https://gitlab.cee.redhat.com/service/managed-tenants

   ```
   export MANAGED_TENANTS_ORIGIN=myuser/managed-tenants
   export MANAGED_TENANTS_FORK=myuser/managed-tenants
   ```

2. Fork the integreatly-operator repo and create two fake releases (tags) or reuse my fork where I've created a fake 2.1.0-rc1 and 2.1.0 releases

   https://github.com/b1zzu/integreatly-operator/releases

   ```
   export INTEGREATLY_ORG=b1zzu
   ```

3. Obtain a GitLab personal token

   https://docs.gitlab.com/ee/user/profile/personal_access_tokens.html

   ```
   export GITLAB_TOKEN=mypersonaltoken
   ```

4. Update the stage project for a build release (2.1.0-rc1)

   ```
   go run main.go release osd-addon \
       --org $INTEGREATLY_ORG \
       --gitlab-token $GITLAB_TOKEN \
       --version 2.1.0-rc1 \
       --managed-tenants-origin $MANAGED_TENANTS_ORIGIN \
       --managed-tenants-fork $MANAGED_TENANTS_FORK
   ```

5. Verify the created MR which will be printed on the console

   ex: https://gitlab.cee.redhat.com/dbizzarr/managed-tenants/merge_requests/1

6. Merge the MR in your fork

7. Update the stage, edge and stable projects for a final release (2.1.0)

   ```
   go run main.go release osd-addon \
       --org $INTEGREATLY_ORG \
       --gitlab-token $GITLAB_TOKEN \
       --version 2.1.0 \
       --managed-tenants-origin $MANAGED_TENANTS_ORIGIN \
       --managed-tenants-fork $MANAGED_TENANTS_FORK
   ```

8. Verify the created MRs which will be printed on the console

   Ex:

   - https://gitlab.cee.redhat.com/dbizzarr/managed-tenants/merge_requests/2
   - https://gitlab.cee.redhat.com/dbizzarr/managed-tenants/merge_requests/3
   - https://gitlab.cee.redhat.com/dbizzarr/managed-tenants/merge_requests/4


cmd/osdAddonRelease.go
```diff
@@ -0,0 +1,552 @@
...
+	}
+
+	return nil
+}
+
+// processCSVImagesCmd represents the processCSVImages command
+var osdAddonReleaseCmd = &cobra.Command{
+	Use:   "osd-addon-release",
+	Short: "crete a release MR for the integreatly-operator to the managed-tenats repo",
+	Run: func(cmd *cobra.Command, args []string) {
```
b1zzu, line 429: I would like to create some tests, but the command is using many dependencies (filesystem, go-git, go-gitlab) so I'm not sure if it makes sense to mock all of these parts

cmd/osdAddonRelease.go
```diff
@@ -0,0 +1,552 @@
...
+		// channels
+		if s, ok := m["channels"].([]interface{}); ok {
+			// [0]
+			if m, ok = s[0].(map[interface{}]interface{}); ok {
+				// .currentCSV
+				m["currentCSV"] = fmt.Sprintf("integreatly-operator.v%s", version)
+				done = true
+			}
+		}
+	}
```
b1zzu, line 255: That's the solution I've found out to edit the yaml file without having to know the full structure and so avoid to remove unknow fields, but maybe I should use a fix structure to make it easier, wdyt?

cmd/osdAddonRelease.go
```diff
@@ -0,0 +1,552 @@
...
+		SourceBranch:    gitlab.String(managedTenantsBranch),
+		TargetBranch:    gitlab.String(managedTenantsMasterBranch),
+		TargetProjectID: gitlab.Int(project.ID),
+	})
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("Merge request for version %s and environment %s created successfully\n", version, channel)
+	fmt.Printf("MR: %s\n", mr.WebURL)
```
b1zzu, line 414: For now I'm just printing out the create MR URL, should I do something more?

cmd/osdAddonRelease.go
```diff
@@ -0,0 +1,552 @@
...
+	gitlabTokenFlag             string
+	mergeRequestDescriptionFlag string
+	managedTenantsOriginFlag    string
+	managedTenantsForkFlag      string
+	integreatlyOperatorFlag     string
+)
+
+// ReleaseChannel rappresents one of the three places (stage, edge, stable)
+// where to update the integreatly-operator
+type ReleaseChannel string
```
wei-lee, line 58: very small thing but does this type needs to be exported?

b1zzu: No, it doesn't I'm just so used to other languages where a type/class/struct always start with an uppercase

cmd/osdAddonRelease.go
```diff
@@ -0,0 +1,552 @@
...
+	if v.build != "" {
+		p = append(p, v.build)
+	}
+	return strings.Join(p, "-")
+}
+
+// IsPreRrelease returns true if the version end with -ER1, -RC1, ...
+func (v *ReleaseVersion) IsPreRrelease() bool {
+	return v.build != ""
+}
```
wei-lee, line 140: It will be good to move these types to "pkg" directory where they can be reused, and reduce the number of code in this file.

cmd/osdAddonRelease.go
```diff
@@ -0,0 +1,552 @@
...
+
+	_, err = io.Copy(out, in)
+	if err != nil {
+		return fmt.Errorf("failed to copy the file content: %s", err)
+	}
+
+	return nil
+}
+
+func copyDirectory(src, dest string) error {
```
wei-lee, line 163: it will be good to move these to it's own file, as I think it's very likely these will be reused.

cmd/osdAddonRelease.go
```diff
@@ -0,0 +1,552 @@
...
+
+	done := false
+	// Set channels[0].currentCSV value
+	if m, ok := i.(map[interface{}]interface{}); ok {
+		// channels
+		if s, ok := m["channels"].([]interface{}); ok {
+			// [0]
+			if m, ok = s[0].(map[interface{}]interface{}); ok {
+				// .currentCSV
+				m["currentCSV"] = fmt.Sprintf("integreatly-operator.v%s", version)
```
wei-lee, line 251: not sure I am a fan of this approach. why not take advantage of the fact that we are using golang, and we can access the CVS golang type definition, and use it to manipulate the csv file. I would imagine we will need to use the CVS types somewhere else sooner or later.

cmd/osdAddonRelease.go
wei-lee: yes, add the relevant operator framework library as a dependency and import the CSV type definition here.

cmd/osdAddonRelease.go
wei-lee: you can definitely add some tests, without mocking for most of the things. Most of the logic here are happening locally (like clone repos, update manifest files, commit changes etc), and you can tests then without mocking, just making sure to clean up when the tests are completed.

The only bit you need to mock is sending PRs to gitlab, which can be done quite easily.

cmd/osdAddonRelease.go
wei-lee: right, in Golang it uses uppercase/lowercase to decide if the object needs to be exported.

cmd/osdAddonRelease.go
```diff
@@ -0,0 +1,552 @@
...
+	"io/ioutil"
+	"os"
+	"path"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/spf13/cobra"
+	"github.com/xanzy/go-gitlab"
+	"gopkg.in/src-d/go-git.v4"
```
wei-lee, line 14: any reason we are not using the latest version: https://github.com/go-git/go-git?

cmd/osdAddonRelease.go
b1zzu: so you are suggesting to rely on github and gitlab to be available for the tests. I can already see a problem because I think the prow cli can not see gitlab so that could be already the first issue

cmd/osdAddonRelease.go
wei-lee: I think that's enough.

cmd/osdAddonRelease.go
b1zzu: Actually is the same package just the v4 version: https://github.com/go-git/go-git/tree/v4.13.1

b1zzu: But it makes sense to update to the v5 version, I tried it and it works in the same way

cmd/osdAddonRelease.go
b1zzu: I moved it to `pkg/utils/copy.go`, is it good?

cmd/osdAddonRelease.go
wei-lee: Yeah, in that case, you can mock git clone implementation to create some files locally, and use them for tests.

cmd/osdAddonRelease.go
```diff
@@ -0,0 +1,462 @@
...
+		&f.version, "version", "",
+		"The RHMI version to push to the managed-tenats repo (ex \"2.0.0\", \"2.0.0-er4\")")
+	cmd.MarkFlagRequired("version")
+
+	cmd.Flags().StringVar(
+		&f.gitlabToken,
+		"gitlab-token",
+		"",
+		"GitLab token to Push the changes and open the MR")
+	cmd.MarkFlagRequired("gitlab-token")
```
wei-lee, line 165: it will be good if this can be provided via the env var or the configuration file (via viper)

cmd/osdAddonRelease.go
```diff
@@ -0,0 +1,462 @@
...
+		err := c.createReleaseMergeRequest(stageChannel)
+		if err != nil {
+			return err
+		}
+
+	} else {
+
+		// When the version is not a prerelease version and is a final release
+		// then create the release against stage, edge and stable
+		for _, channel := range []releaseChannel{stageChannel, edgeChannel, stableChannel} {
```
wei-lee, line 262: can we make the channel configurable via the flags. The reason is that I think we will need to release product versions to stage and edge channels first, then after a while we will release to the stable channel. While we could create all the PRs and leave them open, I worried that someone will merge the PR by accident. Best to create them when they are actually needed.

cmd/osdAddonRelease_test.go
```diff
@@ -0,0 +1,410 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	return dir, repo
+}
+
+func prepareTestEnviornment() {
+
+}
```
wei-lee, line 91: is this used?

cmd/osdAddonRelease_test.go
```diff
@@ -0,0 +1,410 @@
...
+
+	patch, err := fromObject.Patch(toObject)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	return patch
+}
+
+func readFile(t *testing.T, file string) []byte {
```
wei-lee, line 120: this function doesn't seem to be used anywhere.

cmd/osdAddonRelease_test.go
```diff
@@ -0,0 +1,410 @@
...
+			packageManifest := fmt.Sprintf("%s/%s.package.yaml", c.channel.directory(), c.channel.operatorName())
+			clusterServiceVersion := fmt.Sprintf("%s/%s/integreatly-operator.v%s.clusterserviceversion.yaml", c.channel.directory(), version, version)
+			customResourceDefinition := fmt.Sprintf("%s/%s/integreatly.org_rhmis_crd.yaml", c.channel.directory(), version)
+
+			for _, p := range patches {
+				_, file := p.Files()
+				switch file.Path() {
+				case packageManifest:
+					if found := len(p.Chunks()); found != 4 {
+						t.Fatalf("expected 3 but found %d chunk changes for %s", found, packageManifest)
```
wei-lee, line 264: should be`4` in the message?

wei-lee:
@b1zzu might be good to clean up the commits history a bit.

PR #15, opened 2020-04-09
group all release related commands to

wei-lee:
To verify:

1. Build the CLI 
2. run `delorean` and see the help message
3. run `delorean release` and see the help message

### DIFF
--- a/cmd/mergeBlocker.go
+++ b/cmd/mergeBlocker.go
@@ -27,10 +27,8 @@ import (
 
 var baseBranch string
 var isDeletion bool
-var ghOrg string
-var ghRepo string
 
-const MERGE_BLOCKER_LABEL = "tide/merge-blocker"
+const MergeBlockerLabel = "tide/merge-blocker"
 
 // mergeBlockerCmd represents the mergeBlocker command
 var mergeBlockerCmd = &cobra.Command{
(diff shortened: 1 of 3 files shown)


cmd/mergeBlocker.go
```diff
@@ -27,8 +27,6 @@ import (
 
 var baseBranch string
 var isDeletion bool
-var ghOrg string
-var ghRepo string
 
 const MERGE_BLOCKER_LABEL = "tide/merge-blocker"
```
b1zzu, line 31: Not directly related to this PR but what you think about using the Go best practices when writing code?

In this case, the constants should be `mergeBlockerLable` because is not exported:
https://github.com/golang/go/wiki/CodeReviewComments#mixed-caps

Wdyt?

@mikenairn @wei-lee @laurafitzgerald @StevenTobin

cmd/root.go
```diff
@@ -43,18 +52,16 @@ func Execute() {
...
+	//flags for the root command (available for all subcommands)
 	rootCmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file (default is $HOME/.delorean.yaml)")
-	rootCmd.PersistentFlags().StringP("token", "t", "", fmt.Sprintf("Github access token. Can be set via the %s env var.", strings.ToUpper(GITHUB_TOKEN_KEY)))
-	viper.BindPFlag(GITHUB_TOKEN_KEY, rootCmd.PersistentFlags().Lookup("token"))
 
-	// Cobra also supports local flags, which will only run
-	// when this action is called directly.
-	//rootCmd.Flags().BoolP("toggle", "t", false, "Help message for toggle")
+	//flags for the release command (available for all its subcommands)
+	releaseCmd.PersistentFlags().StringP("token", "t", "", fmt.Sprintf("Github access token. Can be set via the %s env var.", strings.ToUpper(GITHUB_TOKEN_KEY)))
```
b1zzu, line 59: Why the token is not set in any variable?

cmd/mergeBlocker.go
wei-lee: ok, will update them

cmd/root.go
```diff
@@ -15,6 +15,8 @@ import (
 )
 
 var cfgFile string
+var ghOrg string
```
b1zzu, line 18: I would use a more extensive name like `integreatlyOperatorOrg` because it may be that we will interact with multiple repos and at that point, it would become confusing

cmd/root.go
wei-lee: we want to get the token value using viper (which will allow us to use environment variables).

PR #17, opened 2020-04-14
Del 222 tag image

wei-lee:
To verify:

1. Make sure you have a valid quy.io api token. Instructions can be found [here](https://docs.quay.io/api/), see "Generating a token" section. You can use this [application](https://quay.io/organization/integreatly/application/A2LWZS1CLEBXG8AKOSQ2?tab=gen-token) to generate a token. Once the token is created, set it via the `QUAY_TOKEN` env var.
2. Fork integreatly-operator into your org
3. Manually create a release PR against your own fork (probably against a test branch rather than master). The release branch name should be something like `release-vx.x.x`
4. Build the cli and run the following:
    ```
    ./delorean release merge-release -o <your-fork> -b <base-branch> -v x.x.x-er1
   ```
   You should see some output and the PR should be merged.
5. To tag the release image (you can test it using the integreatly-operator, just make sure use a tag that hasn't been used, and delete them afterwards from github and quay repos if they get created):
   ```
   ./delorean release tag-release -v 2.0.0-test
   ```
   This will fail if the latest image is not available on quay. You can also try wait for it:
    ```
    ./delorean release tag-release -v 2.0.0-test -w --wait-interval 1
    ```
   You should see some output and it will create the tag once the image is available on quay.


b1zzu:
:eyes:

b1zzu:
Hi @wei-lee,
how can I push the integreatly-operator to my quay org with the commitSHA?

b1zzu:
@wei-lee I've tried to set the commit id manually inside the container but now I see this error:
```
Try to create image tag on quay.io
b1zzu/integreatly-operator fake-master
Error: http error: url = "https://quay.io/api/v1/repository/b1zzu/integreatly-operator/tag/v2.0.0-test"; status = 403
exit status 1
```
probably I have the wrong access token, but how can I generate the right one?

cmd/mergeRelease_test.go
```diff
@@ -0,0 +1,249 @@
...
+		opts        *mergeReleaseOptions
+		expectError bool
+	}{
+		{
+			description: "success",
+			client: &mockPullRequestsService{
+				ListFunc: func(ctx context.Context, owner string, repo string, opts *github.PullRequestListOptions) (requests []*github.PullRequest, response *github.Response, err error) {
+					url := "http://test"
+					num := 1
+					return []*github.PullRequest{&github.PullRequest{
```
b1zzu, line 52:
```suggestion
					return []*github.PullRequest{{
```

cmd/mergeRelease_test.go
```diff
@@ -0,0 +1,249 @@
...
+			opts:        &mergeReleaseOptions{baseBranch: "master", releaseVersion: "2.0.0-rc1"},
+			expectError: true,
+		},
+		{
+			description: "test get PR returns error",
+			client: &mockPullRequestsService{
+				ListFunc: func(ctx context.Context, owner string, repo string, opts *github.PullRequestListOptions) (requests []*github.PullRequest, response *github.Response, err error) {
+					url := "http://test"
+					num := 1
+					return []*github.PullRequest{&github.PullRequest{
```
b1zzu, line 101:
```suggestion
					return []*github.PullRequest{{
```

cmd/tagRelease.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+	fmt.Println("Image tag created:", rv.TagName())
+	return nil
+}
+
+func getGitRef(ctx context.Context, client services.GitService, gitRepoInfo *githubRepoInfo, ref string) (*github.Reference, error) {
+	gitRef, resp, err := client.GetRef(ctx, gitRepoInfo.owner, gitRepoInfo.repo, ref)
+	if err != nil {
+		if resp != nil && resp.StatusCode == http.StatusNotFound {
+			return nil, nil
```
b1zzu, line 101: Shouldn't this line return some error? Or probably add around line 66 an additional if to check that `headRef` is not nill

cmd/tagRelease.go
```diff
@@ -0,0 +1,186 @@
...
+	"github.com/integr8ly/delorean/pkg/quay"
+	"github.com/integr8ly/delorean/pkg/services"
+	"github.com/integr8ly/delorean/pkg/utils"
+	"github.com/spf13/cobra"
+	"net/http"
+	"time"
+)
+
+const (
+	commitIdLabelFilter = "io.openshift.build.commit.id"
```
b1zzu, line 17:
```suggestion
	commitIDLabelFilter = "io.openshift.build.commit.id"
```

cmd/tagRelease.go
```diff
@@ -0,0 +1,186 @@
...
+
+func tryCreateQuayTag(ctx context.Context, quayClient *quay.Client, quayRepo string, existingTag string, newTag string, commitSHA string) error {
+	tags, _, err := quayClient.Tags.List(ctx, quayRepo, &quay.ListTagsOptions{
+		SpecificTag: existingTag,
+	})
+	if err != nil {
+		return err
+	}
+	tag := tags.Tags[0]
+	commitId, _, err := quayClient.Manifests.ListLabels(ctx, quayRepo, *tag.ManifestDigest, &quay.ListManifestLabelsOptions{Filter: commitIdLabelFilter})
```
b1zzu, line 142:
```suggestion
	commitID, _, err := quayClient.Manifests.ListLabels(ctx, quayRepo, *tag.ManifestDigest, &quay.ListManifestLabelsOptions{Filter: commitIDLabelFilter})
```

cmd/tagRelease.go
```diff
@@ -0,0 +1,186 @@
...
+	if err != nil {
+		return err
+	}
+	if *commitId.Labels[0].Value != commitSHA {
```
b1zzu, line 146:
```suggestion
	if *commitID.Labels[0].Value != commitSHA {
```

cmd/tagRelease_test.go
```diff
@@ -0,0 +1,254 @@
...
+	}
+	panic("implement me")
+}
+
+func TestDoTagRelease(t *testing.T) {
+	baseUrl, _ := url.Parse("https://example.com")
+	sha := "testsha"
+	testTagName := "test"
+	testTagDigest := "testdigest"
+	labelKey := commitIdLabelFilter
```
b1zzu, line 70:
```suggestion
	labelKey := commitIDLabelFilter
```

cmd/tagRelease.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+func tryCreateQuayTag(ctx context.Context, quayClient *quay.Client, quayRepo string, existingTag string, newTag string, commitSHA string) error {
+	tags, _, err := quayClient.Tags.List(ctx, quayRepo, &quay.ListTagsOptions{
+		SpecificTag: existingTag,
+	})
+	if err != nil {
+		return err
+	}
+	tag := tags.Tags[0]
```
b1zzu, line 141: Maybe an if checking if there is at least one tag, when I was trying it out, I didn't have any images in my quay repo for the tag and it was panic because Tags len was zero

cmd/tagRelease.go
b1zzu, line 146: Also here the code was panic because I didn't have the `io.openshift.build.commit.id` label set in my image, maybe another if here to check that there is at least one label

cmd/tagRelease.go
```diff
@@ -0,0 +1,186 @@
...
+	tagRef, err := getGitRef(ctx, client, gitRepoInfo, tagRefVal)
+	if err != nil {
+		return nil, err
+	}
+	if tagRef != nil {
+		if tagRef.GetObject().GetSHA() != sha {
+			return nil, fmt.Errorf("tag %s is already created but pointing to a different commit. Please delete it first", tag)
+		} else {
+			return tagRef, nil
+		}
```
b1zzu, line 119:
```suggestion
		}
		return tagRef, nil
```

cmd/tagRelease.go
```diff
@@ -0,0 +1,186 @@
...
+		return fmt.Errorf("can't find an image with given tag %s that matches the given commit SHA: %s", existingTag, commitSHA)
+	} else {
+		_, err := quayClient.Tags.Change(ctx, quayRepo, newTag, &quay.ChangTag{
+			ManifestDigest: *tag.ManifestDigest,
+		})
+		if err != nil {
+			return err
+		}
+		return nil
+	}
```
b1zzu, line 156:
```suggestion
	}
	_, err := quayClient.Tags.Change(ctx, quayRepo, newTag, &quay.ChangTag{
		ManifestDigest: *tag.ManifestDigest,
	})
	if err != nil {
		return err
	}
	return nil
```

cmd/tagRelease.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+func Retry(interval time.Duration, timeout time.Duration, f func() error) error {
```
b1zzu, line 159: I think this function may be useful also for other cmd so I would move it to pkg/ , in alternative we could use the k8s function wait.Poll which is doing almost the same thing: https://github.com/integr8ly/integreatly-operator/blob/e75d38c6442ae42162f3d4398ffd0ed644fcee47/test/common/alerts_firing.go#L71

b1zzu:
I'm seeing a new error:
```
# go run main.go release tag-release -o b1zzu -b fake-master -v 2.0.0-test --quayRepo b1zzupersonal/integreatly-operator

Using config file: /home/b1zzu/.delorean.yaml
Fetch git ref: refs/heads/fake-master
Create git tag: v2.0.0-test
Git tag v2.0.0-test created: https://api.github.com/repos/b1zzu/integreatly-operator/git/refs/tags/v2.0.0-test
Try to create image tag on quay.io
Error: json: cannot unmarshal string into Go value of type quay.Tag
exit status 1
```

cmd/tagRelease.go
wei-lee: I think both `ID` and `Id` are valid.

cmd/tagRelease.go
wei-lee: ok

cmd/tagRelease.go
wei-lee: ok

cmd/tagRelease.go
wei-lee: yeah, I don't want to introduce another dep here.

cmd/tagRelease.go
b1zzu: It's a minor thing but initialisms or acronyms should be al all lowercase (when the are at the beginning) or all uppercase
https://github.com/golang/go/wiki/CodeReviewComments#initialisms

cmd/tagRelease.go
b1zzu: I agree, moving it to pkg/ will be enough

wei-lee:
@b1zzu you comments are addressed now. can you add another `/lgtm` comment if that's ok

PR #19, opened 2020-04-17
update the CSV file before creating MR

wei-lee:
Some extra processing to the CSV file are required when creating the MR to managed-tenants repo, including:

* The `USE_CLUSTER_STORAGE` env var is set to empty string
* The `ALERTING_EMAIL_ADDRESS` env var is added
* The `SingleNamespace` supported namespace is set to true

You should just review this commit: 34a6e123a8a99c6df50ac0fc1358cf23ce912c28. The other one updates the vendor dir.


cmd/osdAddonRelease.go
```diff
@@ -3,21 +3,24 @@ package cmd
 import (
 	"fmt"
 	"io/ioutil"
+	v1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
 	"os"
 	"path"
 	"time"
 
+	"github.com/ghodss/yaml"
```
laurafitzgerald, line 12: Any reason for switching the yaml package. I wasn't sure which one to use. Have been using the remove "gopkg.in/yaml.v2`

cmd/osdAddonRelease.go
```diff
@@ -416,47 +434,141 @@ func (c *osdAddonReleaseCmd) udpateThePackageManifest(channel releaseChannel) (s
...
+	}
+
+	err = yaml.Unmarshal(bytes, obj)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func writeObjectToYAML(obj interface{}, yamlFile string) error {
```
laurafitzgerald, line 511: Could we move this out to a helper? There'll be a need to write files when updating the csv for processing also.

cmd/osdAddonRelease.go
wei-lee: actually yes, apparently the other one (`yaml.v2`) does't parse the CSV file properly, because it is generated using this yaml library. So we need to use the same if we want to parse it.

We can keep them both here, but need to be careful about which one to use.

cmd/osdAddonRelease.go
```diff
@@ -416,47 +434,141 @@ func (c *osdAddonReleaseCmd) udpateThePackageManifest(channel releaseChannel) (s
...
+func findContainerByName(deploymentSpec v1.DeploymentSpec, containerName string) (int, *corev1.Container) {
+	for i, c := range deploymentSpec.Template.Spec.Containers {
+		if c.Name == containerName {
+			return i, &c
+		}
+	}
+	return -1, nil
+}
+
+func addOrUpdateEnvVar(envVars []corev1.EnvVar, envName string, envVal string) []corev1.EnvVar {
```
laurafitzgerald, line 553: Some of these functions would also be useful for processing the manifest command. Also is there an option to push any of those upstream to the operator-sdk?

cmd/osdAddonRelease.go
wei-lee: yeah, will do that.

cmd/osdAddonRelease.go
wei-lee: don't know. I think there might be some functions in the operator-sdk to do this, I just don't know where to find them. It's easy enough to write anyway.

cmd/osdAddonRelease.go
laurafitzgerald: :+1: @wei-lee thanks for the info. I didn't have a preference so can use this one also so it's consistent

cmd/osdAddonRelease.go
laurafitzgerald: ping @mikenairn this is probably something that once moved to utils could be used by your pr. I plan to depend on it for writing the csv file after the updates required after extraction.

cmd/osdAddonRelease.go
laurafitzgerald: 👍 yeah i think it's fine to keep them inside our own code. Could we move it out to utils?

PR #20, opened 2020-04-17
Del 247 add extract manifests cmd

mikenairn:
Add extractManifests command which can extract a manifest bundle from a container image and copy the most recent bundle to a given destination manifest directory and update it's package yaml file.

**Extract manifests from a container image (--src-image)**
```
$ ./delorean ews extract-manifests --src-image registry-proxy.engineering.redhat.com/rh-osbs/amq7-amq-online-1-controller-manager-rhel7-operator-metadata:1.4-36
Extracting manifests from  registry-proxy.engineering.redhat.com/rh-osbs/amq7-amq-online-1-controller-manager-rhel7-operator-metadata:1.4-36
Manifests extracted to  /tmp/manifests-073631170
```
**Extract manifests from a container image (--src-image) and copy the latest bundle to a given manifest directory and update the package.yml default channel (--dest-dir )**
```
$ ./delorean ews extract-manifests --src-image registry-proxy.engineering.redhat.com/rh-osbs/amq7-amq-online-1-controller-manager-rhel7-operator-metadata:1.4-36 --dest-dir /home/mnairn/go/src/github.com/integr8ly/integreatly-operator/manifests/integreatly-amq-online
Extracting manifests from  registry-proxy.engineering.redhat.com/rh-osbs/amq7-amq-online-1-controller-manager-rhel7-operator-metadata:1.4-36
Manifests extracted to  /tmp/manifests-693266346
Copied latest manifest bundle from '/tmp/manifests-693266346/1.4.0' to '/home/mnairn/go/src/github.com/integr8ly/integreatly-operator/manifests/integreatly-amq-online/1.4.0'
```

**Copy the latest bundle from a given maifest directory (--src-dir) to a given manifest directory and update the package.yml default channel (--dest-dir )**

```
$ ./delorean ews extract-manifests --src-dir /tmp/manifests-693266346 --dest-dir /home/mnairn/go/src/github.com/integr8ly/integreatly-operator/manifests/integreatly-amq-online
Copied latest manifest bundle from '/tmp/manifests-693266346/1.4.0' to '/home/mnairn/go/src/github.com/integr8ly/integreatly-operator/manifests/integreatly-amq-online/1.4.0'
```

Related Jiras:

* https://issues.redhat.com/browse/DEL-3

Notes: Still updating tests, but wanted to get eyes on it.

### DIFF
--- a/cmd/extractManifests.go
+++ b/cmd/extractManifests.go
@@ -1,31 +1,116 @@
 package cmd
 
 import (
+	"context"
+	"errors"
 	"fmt"
-
+	"github.com/integr8ly/delorean/pkg/utils"
 	"github.com/spf13/cobra"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
 )
+
+type extractManifestsCmdOptions struct {
+	srcImage string
+	srcDir   string
+	destDir  string
+	tmpDir   string
+}
+
+var extractManifestsCmdOpts = &extractManifestsCmdOptions{}
+
+type ExtractManifestsCmd func(image string, path string) error
+
+func extractImageManifests(image string, destDir string) error {
+	ocExecutable, err := exec.LookPath("oc")
+	if err != nil {
+		return err
+	}
+
+	cmdOCManifestExtract := &exec.Cmd{
+		Path:   ocExecutable,
+		Args:   []string{ocExecutable, "image", "extract", image, "--path", "/manifests/:" + destDir, "--confirm"},
+		Stdout: os.Stdout,
+		Stderr: os.Stdout,
+	}
+
+	return cmdOCManifestExtract.Run()
+}
+
+func copyLatestManifest(srcPkgDir, destPkgDir string) (string, string, error) {
+
+	csv, srcBundleDir, err := utils.GetCurrentCSV(srcPkgDir)
+	if err != nil {
+		return "", "", err
+	}
+
+	destBundleDir := filepath.Join(destPkgDir, filepath.Base(srcBundleDir))
+	err = utils.CopyDirectory(srcBundleDir, destBundleDir)
+	if err != nil {
+		return "", "", err
+	}
+
+	_, err = utils.UpdatePackageManifest(destPkgDir, csv.Name)
+	if err != nil {
+		return "", "", err
+	}
+
+	return srcBundleDir, destBundleDir, nil
+}
+
+func DoExtractManifests(ctx context.Context, extractManifests ExtractManifestsCmd, cmdOpts *extractManifestsCmdOptions) error {
+	if cmdOpts.srcImage == "" && cmdOpts.srcDir == "" {
+		return errors.New("Missing source. Must specify a source image or directory!")
+	}
+	if cmdOpts.srcImage != "" {
+		fmt.Println("Extracting manifests from", cmdOpts.srcImage)
+
+		err := extractManifests(cmdOpts.srcImage, cmdOpts.tmpDir)
+		if err != nil {
+			return err
+		}
+		fmt.Println("Manifests extracted to", cmdOpts.tmpDir)
+		cmdOpts.srcDir = cmdOpts.tmpDir
+	}
+
+	if cmdOpts.destDir != "" {
+		err := utils.VerifyManifestDirs(cmdOpts.srcDir, cmdOpts.destDir)
+		if err != nil {
+			return err
+		}
+
+		from, to, err := copyLatestManifest(cmdOpts.srcDir, cmdOpts.destDir)
+		if err != nil {
+			return err
+		}
+		fmt.Printf("Copied latest manifest bundle from '%s' to '%s'\n", from, to)
+	}
+	return nil
+}
 
 // extractManifestsCmd represents the extractManifests command
 var extractManifestsCmd = &cobra.Command{
 	Use:   "extract-manifests",
 	Short: "Extract OLM Manifest bundle",
-	Long:  `Extracts any olm manifest bundles contained within a given container image and outputs it, unmodified, to a given directory.`,
+	Long:  `Extracts any olm manifest bundles contained within a given source container image or directory and copies the latest bundle, unmodified, to a given directory.`,
 	Run: func(cmd *cobra.Command, args []string) {
-		fmt.Println("extractManifests called")
+		tmpDir, err := ioutil.TempDir(os.TempDir(), "manifests-")
+
+		if err != nil {
+			handleError(err)
+		}
+		extractManifestsCmdOpts.tmpDir = tmpDir
+		if err := DoExtractManifests(cmd.Context(), extractImageManifests, extractManifestsCmdOpts); err != nil {
+			handleError(err)
+		}
 	},
 }
 
 func init() {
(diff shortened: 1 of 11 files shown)


cmd/extractManifests.go
```diff
@@ -2,30 +2,105 @@ package cmd
...
+
+	cmdOCManifestExtract := &exec.Cmd{
+		Path:   ocExecutable,
+		Args:   []string{ocExecutable, "image", "extract", image, "--path", "/manifests/:" + destDir, "--confirm"},
+		Stdout: os.Stdout,
+		Stderr: os.Stdout,
+	}
+
+	return cmdOCManifestExtract.Run()
+}
```
mikenairn, line 29: Just shelling out to the oc binary here, looked into adding oc as a dependency briefly but unsure if that's the best thing to do here or not.

laurafitzgerald:
👀

cmd/extractManifests.go
```diff
@@ -2,30 +2,105 @@ package cmd
...
 	Long:  `Extracts any olm manifest bundles contained within a given container image and outputs it, unmodified, to a given directory.`,
 	Run: func(cmd *cobra.Command, args []string) {
-		fmt.Println("extractManifests called")
+
+		if image != "" {
+			fmt.Println("Extracting manifests from ", image)
+
+			manifestTmpDir, err := ioutil.TempDir(os.TempDir(), "manifests-")
+			if err != nil {
+				return
```
laurafitzgerald, line 69: Print the error here?

cmd/extractManifests.go
```diff
@@ -2,30 +2,105 @@ package cmd
...
+			err = extractImageManifests(image, manifestTmpDir)
+			if err != nil {
+				fmt.Printf("Error extracting manifests from %s: %s", image, err)
+				return
+			}
+			fmt.Println("Manifests extracted to ", manifestTmpDir)
+			srcManifestDir = manifestTmpDir
+		}
+
+		err := utils.VerifyManifestDirs(srcManifestDir, destManifestDir)
```
laurafitzgerald, line 81: This syntax confused me for a sec. Passed two variables. Function accepts an array. Guess it's a golang thing.

cmd/extractManifests.go
```diff
@@ -2,30 +2,105 @@ package cmd
...
+		}
+
 	},
 }
 
 func init() {
+	extractManifestsCmd.Flags().StringVarP(&image, "image", "i", "", "Container image to extract manifests from")
+	extractManifestsCmd.Flags().StringVarP(&srcManifestDir, "src-manifest-dir", "s", "", "Source manifest directory (Ignored if image is provided)")
+	extractManifestsCmd.Flags().StringVarP(&destManifestDir, "dest-manifest-dir", "d", "", "Destination manifest directory")
+	extractManifestsCmd.Flags().BoolVarP(&copyLatest, "copy-latest", "c", true, "Locate and copy the latest manifest bundle from \"src-manifest-dir\" into \"dest-manifest-dir\"")
```
laurafitzgerald, line 103: In what case would this be false?

pkg/utils/csv.go
```diff
@@ -0,0 +1,172 @@
...
+		return nil, "", err
+	}
+
+	var currentCSVName string
+	for _, channel := range pkgManifest.Channels {
+		if channel.IsDefaultChannel(*pkgManifest) {
+			currentCSVName = channel.CurrentCSVName
+			break
+		}
+	}
```
laurafitzgerald, line 119: If we extracted this out to it's own function we could use it seperately to return the currentCSV name which is used for creating the directory inside the integreatly-operator manifests folder

cmd/extractManifests.go
```diff
@@ -2,30 +2,105 @@ package cmd
...
+	"path/filepath"
 )
 
+var image string
+var destManifestDir string
+var srcManifestDir string
+var copyLatest bool
+
+func extractImageManifests(image string, destDir string) error {
+	ocExecutable, _ := exec.LookPath("oc")
```
wei-lee, line 19: might want to deal with the error if `oc` doesn't exist

cmd/extractManifests.go
```diff
@@ -2,30 +2,105 @@ package cmd
...
+	if err != nil {
+		return "", "", err
+	}
+
+	err = utils.UdpatePackageManifest(destPkgDir, csv.Name)
+	if err != nil {
+		return "", "", err
+	}
+
+	err = utils.CopyDirectory(srcBundleDir, destBundleDir)
```
wei-lee, line 49: why we copied the `srcBundleDir` again? will this override the changes we made to the mannifest file?

cmd/extractManifests.go
wei-lee: see https://golang.org/ref/spec#Passing_arguments_to_..._parameters. The compiler will pass them as a slice to the function.

cmd/extractManifests.go
wei-lee, line 69: use [`handleError`](https://github.com/integr8ly/delorean/blob/master/cmd/root.go#L142) function.

mikenairn: Using [`handleError`](https://github.com/integr8ly/delorean/blob/master/cmd/root.go#L142) now.

cmd/extractManifests.go
mikenairn: I have updated the description. I have removed copy-latest entirely and instead made dest-dir optional, if it's set it tries to copy the latest bundle to it. 

I think it will be useful to have the ability to just extract the manifests to a temp dir, for testing as much as anything, i quite often find myself wanting to just check the manifests in a given image.

pkg/utils/csv.go
mikenairn: We could, you could also just use this and get the name from the csv object it returns. If there is some reason to do it, we can.

How would you be using this name to create a directory btw?

cmd/extractManifests.go
mikenairn: :man_shrugging:

cmd/extractManifests.go
mikenairn: :+1:

mikenairn:
@laurafitzgerald 
> Also a general question about the format of the folder copied by the first command. The example provided is 3scale. In the case of the integreatly-operator manifests the folder name is 3scale-x.y.z rather than x.y.z

I'm going to drop the prefix from the bundle directory names. After discussion with Phil, it's not required and we would be better to keep the manifests as similar to the product ones as possible.

> In the first command the folder is extracted from the image and placed in a /tmp/randomstring folder and then to the destination dir (product manifest folder inside integreatly-operator)
In the second case the user can pass an already extracted manifest to the destination dir.
When would a user need the second case given the first case completes that step already? It's mentioned that we want both in the JIRA but I can't recall why.

If we have a product/component that is not productised (UPS/CRO/monitoring), we can use the same command to copy the latest bundle from one of these. We had a Jira about it https://issues.redhat.com/browse/DEL-3

pkg/utils/csv.go
laurafitzgerald: In the case of [3scale](https://github.com/integr8ly/integreatly-operator/tree/master/manifests/integreatly-3scale) the currentCSV is used for the directory names. Same for [amq-online](https://github.com/integr8ly/integreatly-operator/tree/master/manifests/integreatly-amq-online) the dirs are named `amq-online-1.3.1` and `amq-online-1.4.0` currentlt

PR #21, opened 2020-04-21
Modify GetCurrentCSV to return csv file path instead of bundle dir

mikenairn:
/cc @laurafitzgerald


laurafitzgerald:
Building and tests passing.
/lgtm
/approve

PR #22, opened 2020-04-21
processManifest update envs

laurafitzgerald:
**JIRA**
https://issues.redhat.com/browse/DEL-248

NOTE: work in progress but pushing for feedback.

**Verification**
_Build_
`make build/cli`
_Run_
```./delorean ews process-manifest --manifest-dir ./pkg/utils/testdata/validManifests/3scale2/```
_Test_
`cd pkg/utils`
`go test`

### DIFF
--- a/cmd/processManifest.go
+++ b/cmd/processManifest.go
@@ -42,7 +42,11 @@ var processManifestCmd = &cobra.Command{
 		}
 
 		//update "WATCH_NAMESPACE" and "NAMESPACE" env vars if present
-		updateEnvs(csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs)
+		envKeyValMap := map[string]string{
+			envVarWatchNamespace: "metadata.annotations['olm.targetNamespaces']",
+			envVarNamespace:      "metadata.annotations['olm.targetNamespaces']",
+		}
+		utils.UpdateEnvVarList(csv, envKeyValMap)
 
 		//parse object to yaml and write to back to file
 		err = utils.WriteObjectToYAML(csv, filepath)
--- a/cmd/processManifest.go
+++ b/cmd/processManifest.go
@@ -1,9 +1,17 @@
 package cmd
 
 import (
-	"fmt"
-
+	"github.com/integr8ly/delorean/pkg/utils"
+	olmapiv1alpha1 "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
 	"github.com/spf13/cobra"
+	v1 "k8s.io/api/core/v1"
+)
+
+var manifestDir string
+
+const (
+	envVarWatchNamespace = "WATCH_NAMESPACE"
+	envVarNamespace      = "NAMESPACE"
 )
 
 // processManifestCmd represents the processImageManifests command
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

**Comment by laurafitzgerald** [comment]
@mikenairn new pr. I'm adding some failure tests but should be good for a review/verification now.

File: cmd/processManifest.go [hunk 1]
```diff
@@ -12,20 +22,80 @@ var processManifestCmd = &cobra.Command{
...
+		}
+		if filepath == "" {
+			handleError(fmt.Errorf("No csv file found in the directory"))
+		}
+
+		//get yaml from file and parse to object
+		err = utils.PopulateObjectFromYAML(filepath, csv)
+		if err != nil {
+			handleError(err)
+		}
```
mikenairn, line 43: You should be able to remove this since https://github.com/integr8ly/delorean/pull/21

pkg/utils/csv_test.go
```diff
@@ -214,6 +270,13 @@ func TestReadCSVFromBundleDirectory(t *testing.T) {
 				} else {
 					t.Errorf("ReadCSVFromBundleDirectory() got = %v", got)
 				}
+				if got1 != "" {
+					if got1 != tt.want1 {
+						t.Errorf("GetCurrentCSV() got1 = %v, want %v", got1, tt.want1)
+					}
+				} else {
+					t.Errorf("GetCurrentCSV() got1 = %v", got1)
+				}
```
mikenairn, line 279: Looks to be duplicate

laurafitzgerald: Yeah, think we both added it. Didn't spot it in the rebase. Removing now. 👍

cmd/processManifest.go
```diff
@@ -12,20 +22,74 @@ var processManifestCmd = &cobra.Command{
...
-	// processManifestCmd.Flags().BoolP("toggle", "t", false, "Help message for toggle")
+	//update "WATCH_NAMESPACE" and "NAMESPACE" env vars if present
+	for i, env := range envs {
+		if env.Name == envVarWatchNamespace {
+			envs[i] = watchNamespaceEnv
+		}
+		if env.Name == envVarNamespace {
+			envs[i] = namespaceEnv
+		}
+	}
```
laurafitzgerald, line 94: @mikenairn just wanting to call out that this section will only update the env var if it's already present. Based on the `-next` branch that should be fine. e.g. 3scale has WATCH_NAMESPACE env but no NAMESPACE and amq-online has NAMESPACE and no WATCH_NAMESPACE. I played around with updateOrAdd logic here but doesn't look like an env var needs to be added, only updated. Let me know if I'm understanding that incorrectly.

cmd/processManifest.go
```diff
@@ -12,20 +22,74 @@ var processManifestCmd = &cobra.Command{
...
+			handleError(err)
+		}
+		//get current csv object and filepath
+		csv, filepath, err := utils.GetCurrentCSV(manifestDir)
+		if err != nil {
+			handleError(err)
+		}
+		if filepath == "" {
+			handleError(fmt.Errorf("No csv file found in the directory"))
+		}
```
mikenairn, line 37: Is filepath likely to be empty. If no csv file was found i think GetCurrentCSV should throw an error?

cmd/processManifest.go
```diff
@@ -12,20 +22,74 @@ var processManifestCmd = &cobra.Command{
...
+
+		//Get the correct replaces value and update it.
+		sortedCSVs, err := utils.GetSortedCSVNames(manifestDir)
+		if err != nil {
+			handleError(err)
+		}
+
+		if sortedCSVs.Len() > 1 {
+			csv.Spec.Replaces = sortedCSVs[(sortedCSVs.Len() - 2)].Name
+		}
```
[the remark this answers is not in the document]
mikenairn: OK, i looked at it again and i see that you are assuming there will only ever be at most 2 bundles in the directory.

This is true i guess, for the integreatly operator, at the moment, but we could make it less likely to fail if this was to change by searching for the current csv and getting the one before it in the array:

```
var previousCSVName string
for i := range sortedCSVs {
	if sortedCSVs[i].Name == csv.Name {
		if(i >= 1) {
			previousCSVName = sortedCSVs[i-1].Name
			break
		}
	}
}
fmt.Printf("current csv: %s replaces csv: %s\n", csv.Name, previousCSVName)
```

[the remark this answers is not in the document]
laurafitzgerald: The assumption is that the csv that's being updated is the latest version in the list of sorted. This assumption is valid if there are 2 or more. The if is to skip if there is only 1. There would be no replaces is there is only one bundle. So it looks at the array of sorted items and takes the 2nd last item as the replaces. They are sorted by the spec.version semver value.
The loop you have above is more specific tho so would be better if let's say we got an image update which had a version lower than the latest in the integereatly-operator/manifests. Unlinkely but it is more specific. Either will work so maybe better to go with more specific. Will make the change in the morning. 👍

cmd/processManifest.go
laurafitzgerald: Yes, I even skipped testing that inside the GetSortedCSVs method because that's tested elsewhere. Will remove this in the morning. 👍

cmd/processManifest.go
```diff
@@ -12,20 +22,74 @@ var processManifestCmd = &cobra.Command{
...
-	// Here you will define your flags and configuration settings.
+	processManifestCmd.Flags().StringVarP(&manifestDir, "manifest-dir", "m", "", "Manifest Directory Location.")
+}
 
-	// Cobra supports Persistent Flags which will work for this command
-	// and all subcommands, e.g.:
-	// processManifestCmd.PersistentFlags().String("foo", "", "A help for foo")
+func updateEnvs(deployments []olmapiv1alpha1.StrategyDeploymentSpec) {
+	spec := deployments[0].Spec
+	envs := spec.Template.Spec.Containers[0].Env
```
mikenairn, line 68: Are we sure that these are always in the first container in the first deployment?

cmd/processManifest.go
mikenairn: In this case we only need to update. We will eventually need to potentially add env vars to override images, but that's another piece of work.

Would it be more useful to have a more generic method, possibly in the csv utils that takes in a csv and map of env vars to update?

```
envKeyValMap := map[string]string{
	"WATCH_NAMESPACE": "metadata.annotations['olm.targetNamespaces']",
	"NAMESPACE": "metadata.annotations['olm.targetNamespaces']",
}

for _, d := range csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs {
	for _, c := range d.Spec.Template.Spec.Containers {
		for _, e := range c.Env {
			for k, v := range envKeyValMap {
				if e.Name == k {
					e.ValueFrom.FieldRef.FieldPath = v
				}
			}
		}
	}
}
```

cmd/processManifest.go
```diff
@@ -12,20 +22,74 @@ var processManifestCmd = &cobra.Command{
...
+
+		if sortedCSVs.Len() > 1 {
+			csv.Spec.Replaces = sortedCSVs[(sortedCSVs.Len() - 2)].Name
+		}
+
+		//update "WATCH_NAMESPACE" and "NAMESPACE" env vars if present
+		updateEnvs(csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs)
+
+		//parse object to yaml and write to back to file
+		err = utils.WriteObjectToYAML(csv, filepath)
```
mikenairn, line 53: When this is written back to file the contents are very different form the original. I don't suppose there is anything we can do about that doing it this way?

laurafitzgerald: Yeah I saw that alright. The issue is with Marshall and Unmarshal. the yaml package orders it in some way, in this case alphabetically. There's mention of it here: https://blog.labix.org/2014/09/22/announcing-yaml-v2-for-go#mapslice They introdced a MapSlice and MapItem to try to solve it. I'm looking at that now to see if we can use it. @wei-lee did you see this when updating csvs in your changes?

cmd/processManifest.go
laurafitzgerald: As they are operator deployments and there's usually one operator I have made an assumption here alright. That of course could change so there might be no harm in making this more specific.

PR #24, opened 2020-04-22
DEL-224 Fix PR!13 open comments

b1zzu:
Follow up: https://github.com/integr8ly/delorean/pull/13

* Bind the `gitlab-token` flag to viper so that it could be set through env and delorean config file
* Remove unsued functions
* Make the release channel configurable through flags

### DIFF
--- a/cmd/osdAddonRelease.go
+++ b/cmd/osdAddonRelease.go
@@ -119,6 +119,7 @@ func gitCloneToTmp(prefix string, url string, reference plumbing.ReferenceName) 
 
 type osdAddonReleaseFlags struct {
 	version                 string
+	channel                 string
 	mergeRequestDescription string
 	managedTenantsOrigin    string
 	managedTenantsFork      string
@@ -128,6 +129,7 @@ type osdAddonReleaseCmd struct {
 	flags                  *osdAddonReleaseFlags
 	gitlabToken            string
 	version                *utils.RHMIVersion
+	channel                releaseChannel
 	gitlabMergeRequests    services.GitLabMergeRequestsService
 	gitlabProjects         services.GitLabProjectsService
 	integreatlyOperatorDir string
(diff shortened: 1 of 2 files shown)


cmd/osdAddonRelease.go
```diff
@@ -253,100 +271,83 @@ func newOSDAddonReleseCmd(flags *osdAddonReleaseFlags) (*osdAddonReleaseCmd, err
...
-		// then create the release against stage, edge and stable
-		for _, channel := range []releaseChannel{stageChannel, edgeChannel, stableChannel} {
-			err := c.createReleaseMergeRequest(channel)
-			if err != nil {
-				return err
-			}
+	// verify that the passed version can be pushed to the passed channel
+	switch c.channel {
+	case stageChannel:
+		// noting
```
mikenairn, line 277: Is this doing anything?

cmd/osdAddonRelease.go
```diff
@@ -140,8 +146,14 @@ func init() {
 		Short: "Create the integreatly-operator MR to the managed-tenants repo",
 		Run: func(cmd *cobra.Command, args []string) {
 
+			gitlabToken, err := requireToken(gitlabTokenKey)
+			if err != nil {
+				fmt.Printf("error: %s\n", err)
+				os.Exit(1)
+			}
```
mikenairn, line 153: Should we be using handleError here to keep things consistent? https://github.com/integr8ly/delorean/blob/master/cmd/mergeBlocker.go#L30

mikenairn:
@wei-lee Would you mind looking over these chnages when you can, i think you were more familiar with the asks here.

cmd/osdAddonRelease.go
b1zzu: is doing nothing, is there to avoid to enter in  `default` if the channel is stage, if it's confusing I can change the switch with an if/else

cmd/osdAddonRelease.go
b1zzu: yes, change it now

cmd/osdAddonRelease.go
mikenairn: Ah sorry, i thought the case below was also checking `stageChannel` , but it's `stableChannel`

PR #27, opened 2020-04-22
Add OWNERS file to make and ocm dirs and added psturc as approver

mikenairn:
/cc @psturc


psturc:
thanks @mikenairn - still someone else than me needs to approve this PR.

PR #28, opened 2020-04-22
add create-release command

wei-lee:
You can see an example PR created by this command: https://github.com/wei-lee/integreatly-operator/pull/5.

To verify:

1. Fork the integr8ly/integreatly-operator repo if you haven't done so. Create a new branch that is based off this one: https://github.com/StevenTobin/integreatly-operator/tree/DEL-219.
2. Generate a github personal access token if you haven't done so.
3. Set the following env vars (best to create a `.delorean.yaml` config file in your home dir and add the following env vars there):
    ```
    export GITHUB_USER=<your github username>
    export GITHUB_TOKEN=<your github access token>
    ```
4. Run the command:
    ```
     ./delorean release create-release --owner <your-fork> --repo integreatly-operator --baseBranch <your-base-branch> --version 2.2.0-rc1
    ```
5. You should see some output and when the command finished, a PR is created.


b1zzu:
:eyes:

cmd/createRelease.go
```diff
@@ -0,0 +1,219 @@
...
+	}, nil
+}
+
+func (c *createReleaseCmd) run(ctx context.Context) (string, error) {
+	fmt.Println(fmt.Sprintf("Clone repo from %s/%s/%s.git to a temporary directory", githubURL, c.repoInfo.owner, c.repoInfo.repo))
+	repoDir, gitRepo, err := c.gitCloneService.CloneToTmpDir("integreatly-operator", fmt.Sprintf("%s/%s/%s.git", githubURL, c.repoInfo.owner, c.repoInfo.repo), c.baseBranch)
+	if err != nil {
+		return "", err
+	}
+	fmt.Println(fmt.Sprintf("Repo cloned to %s", repoDir))
```
b1zzu, line 103: why not just using `fmt.Printf`?

cmd/createRelease.go
```diff
@@ -0,0 +1,219 @@
...
+		gitUser:         user,
+		gitPass:         token,
+		gitCloneService: &services.DefaultGitCloneService{},
+		gitPushService:  &services.DefaultGitPushService{},
+	}, nil
+}
+
+func (c *createReleaseCmd) run(ctx context.Context) (string, error) {
+	fmt.Println(fmt.Sprintf("Clone repo from %s/%s/%s.git to a temporary directory", githubURL, c.repoInfo.owner, c.repoInfo.repo))
+	repoDir, gitRepo, err := c.gitCloneService.CloneToTmpDir("integreatly-operator", fmt.Sprintf("%s/%s/%s.git", githubURL, c.repoInfo.owner, c.repoInfo.repo), c.baseBranch)
```
b1zzu, line 99: it would be good to print the tmp dir as it would be useful for debugging failures

cmd/createRelease.go
```diff
@@ -0,0 +1,219 @@
...
+		return "", err
+	}
+	fmt.Println(fmt.Sprintf("Repo cloned to %s", repoDir))
+
+	gitRepoTree, err := gitRepo.Worktree()
+	if err != nil {
+		return "", err
+	}
+
+	releaseBranchName := fmt.Sprintf("prepare-for-release-%s", c.version.TagName())
```
b1zzu, line 110: maybe using a constant for the release branch name template so that it can be reused easily

cmd/root.go
```diff
@@ -74,6 +75,8 @@ func init() {
 	//flags for the release command (available for all its subcommands)
 	releaseCmd.PersistentFlags().StringP("token", "t", "", fmt.Sprintf("Github access token. Can be set via the %s env var.", strings.ToUpper(GithubTokenKey)))
 	viper.BindPFlag(GithubTokenKey, releaseCmd.PersistentFlags().Lookup("token"))
+	releaseCmd.PersistentFlags().StringP("user", "u", "", fmt.Sprintf("Github user. Can be set via the %s env var.", strings.ToUpper(GithubUserKey)))
```
b1zzu, line 78: For pushing the user is not needed, it can be everything because github will only use the token to authenticate the user, I think it simplifies the command if we remove it and substitute with a fake username

pkg/services/git_service.go
```diff
@@ -3,10 +3,46 @@ package services
...
 	Push(o *git.PushOptions) error
 }
+
+type GitCloneService interface {
+	CloneToTmpDir(prefix string, url string, reference plumbing.ReferenceName) (string, *git.Repository, error)
+}
+
+type GitPushService interface {
+	Push(gitRepo *git.Repository, opts *git.PushOptions) error
+}
```
b1zzu, line 22: Can I ask why creating the `GitPushService`? Can't we reuse the `GitRepositoryService`?

cmd/osdAddonRelease.go
```diff
@@ -221,8 +198,9 @@ func newOSDAddonReleseCmd(flags *osdAddonReleaseFlags, gitlabToken string) (*osd
 	}
 	fmt.Print("gitlab client initialized and authenticated\n")
 
+	gitCloneService := &services.DefaultGitCloneService{}
 	// Clone the managed tenants
-	managedTenatsDir, managedTenantsRepo, err := gitCloneToTmp(
+	managedTenatsDir, managedTenantsRepo, err := gitCloneService.CloneToTmpDir(
```
b1zzu, line 203:
```suggestion
	// TODO: Move the clone functions inise the run() method to improve the test covered code
	managedTenatsDir, managedTenantsRepo, err := gitCloneService.CloneToTmpDir(
```
I would like to add this TODO so the next time we put hands on the osdAddonRelease we can do it

cmd/createRelease.go
wei-lee: it is the `repoDir` and it's printed out in L103.

cmd/root.go
wei-lee: it looks like user is required according to their docs: https://help.github.com/en/github/authenticating-to-github/creating-a-personal-access-token-for-the-command-line#using-a-token-on-the-command-line

cmd/createRelease.go
b1zzu: my bad didn't saw it

cmd/root.go
b1zzu: I set it to `fake` and it's still working, same was for gitlab

pkg/services/git_service.go
wei-lee: We can, but then I have to implement other methods that I don't actually need. If you think about it, we never actually need to override the `Head` and `Worktree` method, it's only the `Push` method that we will need to mock. So rather than reuse the existing `GitRepositoryService`, I think it's better to remove the push method from it, and use the `GitPushService` instead. I will update it.

cmd/root.go
wei-lee: Hmm, but it is not documented, so not sure if this will keep working in the future though. I think I'd rather keep it this way for now, it is very little overhead to set it, and you actually just need to set it once in the config file.

pkg/services/git_service.go
wei-lee: I refactored it a bit here. As you can see, we can now get rid of the `GitRepositoryService` completely now.

wei-lee:
@b1zzu I have addressed most of your comments. can you add the `lgtm` label again if you are ok. then I can merge it and close the PR. Thanks.

pkg/services/git_service.go
b1zzu: If we get rid of the `GitRepositoryService` it's also fine, I just want to have a single solution

PR #30, opened 2020-04-24
make sure tag the test harness image, and also filter the git tag ref…

wei-lee:
…s when list them from github api

The PR fix a few small issues:

1. The `tagRelease` command was failing when there are multiple tags that have the same base but different builds.
2. Make sure the `tagRelease` command can tag multiple quay images (integreatly-operator and test-harness image). If nothing is passed, it will skip the step to tag quay images (easier for demo).
3. Fix an issue when copy the OLM manifest file when run the `osd-addon` command.
4. Make sure the right email address is used for staging and production enviornments.


cmd/osdAddonRelease.go
```diff
@@ -48,6 +48,7 @@ const (
 	envVarNameUseClusterStorage  = "USE_CLUSTER_STORAGE"
 	envVarNameAlerEmailAddress   = "ALERTING_EMAIL_ADDRESS"
 	integreatlyAlertEmailAddress = "integreatly-notifications@redhat.com"
+	cssreAlertEmailAddress       = "cssre-alerts@redhat.com"
```
mikenairn, line 51: Do we want to hard code our email addresses in here?

mikenairn:
One question, not a big one though, will leave it to you tt approve if you are happy with it.

/lgtm

cmd/osdAddonRelease.go
wei-lee: yes, it should be ok. Apparently it's always the same value for production environments. We can make it configurable in the future if required.

PR #36, opened 2020-04-29
Add Dockerfile

mikenairn:
Add Dockerfile
* install oc
* install operator-sdk
* install delorean

Add basic test script, verify it works with:

```
make image/test
```

Jira:
* https://issues.redhat.com/browse/DEL-194

Related PRs:
* https://github.com/openshift/release/pull/8659

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,23 @@
+FROM registry.svc.ci.openshift.org/openshift/release:golang-1.13
+LABEL maintainer="mnairn@redhat.com"
+
+ENV OPERATOR_SDK_VERSION=v0.15.1 \
+    OC_VERSION="4.5" \
+    GOFLAGS=""
+
+# install oc
+RUN curl -Ls https://mirror.openshift.com/pub/openshift-v4/clients/oc/$OC_VERSION/linux/oc.tar.gz | tar -zx && \
+    mv oc /usr/local/bin
+
+# install operator-sdk (from git with no history and only the tag)
+RUN mkdir -p $GOPATH/src/github.com/operator-framework \
+    && cd $GOPATH/src/github.com/operator-framework \
+    && git clone --depth 1 -b $OPERATOR_SDK_VERSION https://github.com/operator-framework/operator-sdk \
+    && cd operator-sdk \
+    && go mod vendor \
+    && make tidy \
+    && make install \
+    && chmod -R 0777 $GOPATH \
+    && rm -rf $GOPATH/.cache
+
+COPY delorean /usr/bin/delorean
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,15 +9,9 @@ ENV OPERATOR_SDK_VERSION=v0.15.1 \
 RUN curl -Ls https://mirror.openshift.com/pub/openshift-v4/clients/oc/$OC_VERSION/linux/oc.tar.gz | tar -zx && \
     mv oc /usr/local/bin
 
-# install operator-sdk (from git with no history and only the tag)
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+    && cd $GOPATH/src/github.com/operator-framework \
+    && git clone --depth 1 -b $OPERATOR_SDK_VERSION https://github.com/operator-framework/operator-sdk \
+    && cd operator-sdk \
+    && go mod vendor \
+    && make tidy \
+    && make install \
+    && chmod -R 0777 $GOPATH \
+    && rm -rf $GOPATH/.cache
+
+COPY delorean /usr/bin/delorean
```
mikenairn, line 23: We could have had an entry point set here to delorean, but didn't think this would work if we want to use it Jenkins:

```
ENTRYPOINT ["/usr/bin/delorean"]
```

Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+# install operator-sdk (from git with no history and only the tag)
+RUN mkdir -p $GOPATH/src/github.com/operator-framework \
+    && cd $GOPATH/src/github.com/operator-framework \
+    && git clone --depth 1 -b $OPERATOR_SDK_VERSION https://github.com/operator-framework/operator-sdk \
+    && cd operator-sdk \
+    && go mod vendor \
+    && make tidy \
+    && make install \
+    && chmod -R 0777 $GOPATH \
+    && rm -rf $GOPATH/.cache
```
b1zzu, line 21: Why downloading the precompiled binary? 
```
https://github.com/operator-framework/operator-sdk/releases/download/${OPERATOR_SDK_VERSION}/operator-sdk-${OPERATOR_SDK_VERSION}-x86_64-linux-gnu
```

Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM registry.svc.ci.openshift.org/openshift/release:golang-1.13
```
b1zzu, line 1: If we download the binary instead of compiling the SDK we would be able to use a smaller container as base

README.md
```diff
@@ -16,6 +16,24 @@ A binary will be created in the root directory of the repo, which can be run:
...
+To run some basic tests against the image:
+
+```
+make image/test
+```
+
+To build and test using a container engine other than docker (podman):
+
+```
+make image/test CONTAINER_ENGINE=podman
```
b1zzu, line 34: +1 for podman support

scripts/image/test
```diff
@@ -0,0 +1,26 @@
+#!/bin/sh
+
+set -e
```
b1zzu, line 3:
```suggestion
set -ex
```
it would make the test easier to debug if it fails in the CI

Dockerfile
mikenairn: Yeah could do, i have no idea why we started doing it this way tbh, copied this from another Dockerfile we use.

scripts/image/test
mikenairn: updated

Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM centos:7
```
mikenairn, line 1: I changed the base image to centos:7 since we don't need to compile any go in here.

make/image.mk
```diff
@@ -0,0 +1,14 @@
+
+CONTAINER_ENGINE ?= docker
+
+DELOREAN_IMAGE=quay.io/integreatly/delorean-cli:latest
```
mikenairn, line 4: fyi, using "delorean-cli" as we were already using "delorean" for something else.

PR #37, opened 2020-04-29
Update test/unit to run ./pkg tests

mikenairn:
/cc @wei-lee 

Some quay tests are failing, could you take a look.


wei-lee:
replaced by #38

PR #38, opened 2020-04-30
run tests in the pkg/ directory

wei-lee:
replace #37

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ test/lint:
 .PHONY: lint
 
 test/unit:
-	go test -v ./cmd ./pkg/*
+	go test -v ./cmd ./pkg/...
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -14,7 +14,7 @@ test/lint:
 .PHONY: lint
 
 test/unit:
-	go test -v ./cmd
+	go test -v ./cmd ./pkg/*
```
b1zzu, line 17: Using the `./pkg/*` will only run the test on the first level like `./pkg/utils`, but if we add another level like `./pkg/utils/foo` then they will be not executed, the correct way to recursive run all tests in a directory should be `./pkg/...`
```suggestion
	go test -v ./cmd ./pkg/...
```

PR #39, opened 2020-04-30
Move Dockerfile to build directory

mikenairn:


mikenairn:
/cc @wei-lee 

image job will fail until i update CI

PR #40, opened 2020-04-30
if an image tag is specified when run tag-release, use it as the quay…

wei-lee:
… image tag

### DIFF
--- a/cmd/tagRelease.go
+++ b/cmd/tagRelease.go
@@ -147,7 +147,8 @@ func tryCreateQuayTag(ctx context.Context, quayClient *quay.Client, quayRepos st
 	ok := true
 	for _, r := range repos {
 		fmt.Println("Create image tag for", r)
-		err := createTagForImage(ctx, quayClient, r, existingTag, newTag, commitSHA)
+		repo, tag := getImageRepoAndTag(r, newTag)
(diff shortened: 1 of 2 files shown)


wei-lee:
@mikenairn this will allow us to update the `latest-staging` tag for the test harness repo. all we need to do then is to add `integreatly/integreatly-operator-test-harness:latest-staging` to the flags when run the `tag-release` command.

mikenairn:
@wei-lee This is all going to happen when we are waiting for the images to sync to quay.io right? When we are creating the release?

I think the "latest-staging" tag would need to be set after the Merge Request to update the staging config is merged otherwise we will be updating the test image before staging is updated.

wei-lee:
@mikenairn given that this will be running in a pipeline, in most cases, there shouldn't be a problem to update the tag when the release is cut. If there is a problem, someone can always revert the tag - it still better than manually update the tag every time?

mikenairn:
/lgtm

This doesn't stop us changing how we do it in the pipeline so approving.

I think however we might be able to just add an extra step in the pipeline that does this at the correct time, we don't really know how long those MRs might take to merge, things happen.

Updating the image can be done fairly easily using oc:

```
oc image mirror quay.io/integreatly/integreatly-operator-test-harness:v2.1.0 quay.io/integreatly/integreatly-operator-test-harness:latest-staging
```

PR #47, opened 2020-05-07
Del 254 add process csv images

StevenTobin:
Add process-csv-images command.

To Verify
- Checkout this branch
- `Make build/cli`
- `./delorean ews process-csv-images --manifest-dir ../integreatly-operator/manifests/integreatly-<product>/`
- For fuse use the command `/delorean ews process-csv-images --manifest-dir ../integreatly-operator/manifests/integreatly-fuse-online/ -e DV_IMAGE=registry.redhat.io/fuse7-tech-preview/fuse-dv-rhel7:1.6 -e OAUTH_IMAGE=registry.redhat.io/openshift4/ose-oauth-proxy:4.2 -e UI_IMAGE=registry.redhat.io/fuse7/fuse-ignite-ui:1.6 -e S2I_IMAGE=registry.redhat.io/fuse7/fuse-ignite-s2i:1.6 -e PROMETHEUS_IMAGE=registry.redhat.io/openshift3/prometheus:v3.9 -e UPGRADE_IMAGE=registry.redhat.io/fuse7/fuse-ignite-upgrade:1.6 -e META_IMAGE=registry.redhat.io/fuse7/fuse-ignite-meta:1.6 -e PSQL_EXPORTER_IMAGE=registry.redhat.io/fuse7-tech-preview/fuse-postgres-exporter:1.6 -e SERVER_IMAGE=registry.redhat.io/fuse7/fuse-ignite-server:1.6` to include the extra image vars the operator needs
- Verify the csv images have been changed to point to `quay.io/integreatly/delorean...` and that the image_mirror_mapping exists and is populated

TODO

- Image entries can be duplicated, should dedupe before writing to the image_mirror_mapping file

### DIFF
--- a/cmd/processCSVImages.go
+++ b/cmd/processCSVImages.go
@@ -1,32 +1,80 @@
 package cmd
 
 import (
-	"fmt"
-
+	"context"
+	"errors"
+	"github.com/integr8ly/delorean/pkg/utils"
 	"github.com/spf13/cobra"
+	"os"
+	"path"
 )
+
+type processCSVImagesCmdOptions struct {
+	manifestDir string
+	isGa        bool
+	extraImages []string
+}
+
+type ProcessCSVImagesCmd func(manifestDir string, isGa bool, extraImages string) error
+
+var processCSVImagesCmdOpts = &processCSVImagesCmdOptions{}
 
 // processCSVImagesCmd represents the processCSVImages command
 var processCSVImagesCmd = &cobra.Command{
 	Use:   "process-csv-images",
 	Short: "Replace internal image registry references and generates an image mirror mapping file.",
 	Long: `Locates the current cluster service version file (csv) for a given product and replaces all occurrences of 
-internal image registries with a deloeran version and generates an image_mirror_mapping file.`,
+internal image registries with a delorean version and generates an image_mirror_mapping file.`,
 	Run: func(cmd *cobra.Command, args []string) {
-		fmt.Println("processCSVImages called")
+		err := DoProcessCSV(cmd.Context(), processCSVImagesCmdOpts)
+		if err != nil {
+			handleError(err)
+		}
+
 	},
+}
+
+func DoProcessCSV(ctx context.Context, cmdOpts *processCSVImagesCmdOptions) error {
+	if cmdOpts.manifestDir == "" {
+		return errors.New("manifest-dir not specified")
+	}
+
+	//verify it's a manifest dir.
+	err := utils.VerifyManifestDirs(cmdOpts.manifestDir)
+	if err != nil {
+		handleError(err)
+	}
+
+	if !cmdOpts.isGa {
+		images, err := utils.GetAndUpdateOperandImagesToDeloreanImages(cmdOpts.manifestDir, cmdOpts.extraImages)
+		images, err = utils.UpdateOperatorImagesToDeloreanImages(cmdOpts.manifestDir, images)
+		if err != nil {
+			handleError(err)
+		}
+		if len(images) > 0 {
+			err = utils.CreateImageMirrorMappingFile(cmdOpts.manifestDir, images)
+			if err != nil {
+				handleError(err)
+			}
+		}
+	}
+
+	if cmdOpts.isGa {
+		if utils.FileExists(path.Join(cmdOpts.manifestDir, utils.MappingFile)) {
+			err := os.Remove(path.Join(cmdOpts.manifestDir, utils.MappingFile))
+			if err != nil {
+				handleError(err)
+			}
+		}
+	}
+
+	return nil
 }
 
 func init() {
 	ewsCmd.AddCommand(processCSVImagesCmd)
 
-	// Here you will define your flags and configuration settings.
-
-	// Cobra supports Persistent Flags which will work for this command
-	// and all subcommands, e.g.:
-	// processCSVImagesCmd.PersistentFlags().String("foo", "", "A help for foo")
-
-	// Cobra supports local flags which will only run when this command
-	// is called directly, e.g.:
-	// processCSVImagesCmd.Flags().BoolP("toggle", "t", false, "Help message for toggle")
+	processCSVImagesCmd.Flags().StringVarP(&processCSVImagesCmdOpts.manifestDir, "manifest-dir", "m", "", "Manifest Directory Location.")
+	processCSVImagesCmd.Flags().BoolVarP(&processCSVImagesCmdOpts.isGa, "isGa", "g", false, "Mark as GA version")
+	processCSVImagesCmd.Flags().StringArrayVarP(&processCSVImagesCmdOpts.extraImages, "extra-images", "e", []string{}, "Extra images to include in container env NAME@IMAGE")
 }
--- a/cmd/processCSVImages.go
+++ b/cmd/processCSVImages.go
@@ -52,7 +52,7 @@ func DoProcessCSV(ctx context.Context, cmdOpts *processCSVImagesCmdOptions) erro
 			handleError(err)
 		}
 		if len(images) > 0 {
-			err = utils.CreateImageMirrorMappingFile(cmdOpts.manifestDir, images)
+			err = utils.WriteToFile(path.Join(cmdOpts.manifestDir, utils.MappingFile), images)
 			if err != nil {
 				handleError(err)
 			}
--- a/pkg/utils/csv.go
+++ b/pkg/utils/csv.go
@@ -3,15 +3,27 @@ package utils
 import (
 	"fmt"
 	"io/ioutil"
+	corev1 "k8s.io/api/core/v1"
 	"os"
 	"path"
 	"path/filepath"
+	"regexp"
 	"sort"
 	"strings"
 
 	"github.com/blang/semver"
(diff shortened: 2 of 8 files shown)


pkg/utils/io.go
```diff
@@ -94,3 +94,32 @@ func deleteKeyFromUnstructured(u map[string]interface{}, key string) {
...
+	info, err := os.Stat(filename)
+	if os.IsNotExist(err) {
+		return false
+	}
+	return !info.IsDir()
+}
+
+func CreateImageMirrorMappingFile(manifestDir string, images []string) error {
+
+	f, err := os.Create(manifestDir + "image_mirror_mapping")
```
b1zzu, line 108: I tried to execute the cmd in this way:
```
./delorean ews process-csv-images --manifest-dir ../integreatly-operator/manifests/integreatly-3scale
```
and because I haven't put the slash at the end of the directory this is the file that has been created:
```
manifests/integreatly-3scaleimage_mirror_mapping
```
instead of 
```
manifests/integreatly-3scale/image_mirror_mapping
```
it should be possible to solve this problem by using the `path.Join` [1] method because it will add the `/` between the path if necessary 

1. https://golang.org/pkg/path/#Join

b1zzu, line 108: As the "image_mirror_mapping" string is used in different files I think we should move it to a global constant in `utils/csv.go`

cmd/processCSVImages.go
```diff
@@ -1,32 +1,80 @@
...
+	err := utils.VerifyManifestDirs(cmdOpts.manifestDir)
+	if err != nil {
+		handleError(err)
+	}
 
-	// Cobra supports Persistent Flags which will work for this command
-	// and all subcommands, e.g.:
-	// processCSVImagesCmd.PersistentFlags().String("foo", "", "A help for foo")
+	if !cmdOpts.isGa {
+		images, err := utils.GetAndUpdateOperandImagesToDeloreanImages(cmdOpts.manifestDir, cmdOpts.isGa, cmdOpts.extraImages)
```
b1zzu, line 48: Why passing `isGa`?  It seems to me that it will be always `false` and is not even used in the function

cmd/processCSVImages.go
```diff
@@ -1,32 +1,80 @@
...
+			}
+		}
+
+		err = utils.UpdateOperatorImagesToDeloreanImages(cmdOpts.manifestDir, cmdOpts.isGa)
+		if err != nil {
+			handleError(err)
+		}
+	}
+
+	if cmdOpts.isGa {
```
b1zzu, line 62: Either way doesn't change much, but why not using an `else {`

cmd/processCSVImages.go
```diff
@@ -1,32 +1,80 @@
...
+		if utils.FileExists(cmdOpts.manifestDir + "image_mirror_mapping") {
```
b1zzu, line 63: I would likely add a log message here before the if to state the action that will be performed

pkg/utils/csv.go
```diff
@@ -163,3 +175,150 @@ func UpdatePackageManifest(packageDir, currentCSVName string) (*registry.Package
...
+
+	if len(extraImages) > 0 {
+		csv, err = includeImages(extraImages, csv)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	var images []string
+	_, deployment := FindDeploymentByName(csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs, csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs[0].Name)
```
b1zzu, line 193: Sorry, I don't get why using *FindDeploymentByName* function, maybe I'm missing something but to me this line could be rewritten like this
```golang
deployment := csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs[0]
```

pkg/utils/csv.go
```diff
@@ -163,3 +175,150 @@ func UpdatePackageManifest(packageDir, currentCSVName string) (*registry.Package
...
+	}
+	return images, nil
+}
+
+func includeImages(extraImages []string, csv *olmapiv1alpha1.ClusterServiceVersion) (*olmapiv1alpha1.ClusterServiceVersion, error) {
+	_, deployment := FindDeploymentByName(csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs, csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs[0].Name)
+	if deployment != nil {
+		for n, container := range deployment.Spec.Template.Spec.Containers {
+			for _, i := range extraImages {
+				currImage := strings.Split(i, "@")
```
StevenTobin, line 239: I'm going to change this to split on `=` I think that makes more sense

pkg/utils/csv.go
```diff
@@ -163,3 +175,150 @@ func UpdatePackageManifest(packageDir, currentCSVName string) (*registry.Package
...
+	return csv, nil
+}
+
+func UpdateOperatorImagesToDeloreanImages(manifestDir string, isGa bool) error {
+	csv, fp, err := GetCurrentCSV(manifestDir)
+	if err != nil {
+		return err
+	}
+
+	_, deployment := FindDeploymentByName(csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs, csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs[0].Name)
```
b1zzu, line 255: Here too, isn't the same thing:
```
deployment := csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs[0]
```

pkg/utils/io.go
```diff
@@ -94,3 +94,32 @@ func deleteKeyFromUnstructured(u map[string]interface{}, key string) {
...
+
+func FileExists(filename string) bool {
+	info, err := os.Stat(filename)
+	if os.IsNotExist(err) {
+		return false
+	}
+	return !info.IsDir()
+}
+
+func CreateImageMirrorMappingFile(manifestDir string, images []string) error {
```
b1zzu, line 106: Maybe we could generalize more this function, something like:
```
func WriteStringArray(array []string, file string) error  {
```
and call it in this way
```
err = utils.CreateImageMirrorMappingFile(images, path.Join(cmdOpts.manifestDir, "image_mirror_mapping"))
```

pkg/utils/io.go
StevenTobin: Cool, yeah i'll switch it to path.join to catch that

StevenTobin: :+1:

cmd/processCSVImages.go
StevenTobin: Ah yeah, this is leftover, I used that var before I rewrote the code a bit, i'll remove it.

cmd/processCSVImages.go
StevenTobin: No real reason just safer to actually check the var rather than relying on else

pkg/utils/csv.go
StevenTobin: I don't want to assume that the first container is the one we want so I'm using that `FindDeploymentByName` function to ensure we are only affecting the operator container

pkg/utils/csv.go
StevenTobin: And same as above, I don't want to assume the first container is the one we want to change

cmd/processCSVImages.go
b1zzu: Both ways are fine with me, I'm just used to si if else in this cases :)

pkg/utils/csv.go
b1zzu: But if you pass to the `FindDeploymentByName` the array `csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs` and you ask the function to search for the name `csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs[0].Name`, which is the name of the first element in the same array that is passed, wouldn't this mean that it would always match the first entry in the array

StevenTobin: Ah you're right, i'm already making that assumption. Yeah the change makes sense then.

cmd/processCSVImages.go
```diff
@@ -45,23 +45,23 @@ func DoProcessCSV(ctx context.Context, cmdOpts *processCSVImagesCmdOptions) erro
...
 
-		err = utils.UpdateOperatorImagesToDeloreanImages(cmdOpts.manifestDir, cmdOpts.isGa)
+		err = utils.UpdateOperatorImagesToDeloreanImages(cmdOpts.manifestDir)
 		if err != nil {
 			handleError(err)
 		}
 	}
 
 	if cmdOpts.isGa {
-		if utils.FileExists(cmdOpts.manifestDir + "image_mirror_mapping") {
```
b1zzu: We should use path.Join here too otherwise we could have the same issue if the manifestDir is not passing with the / at the end

StevenTobin: Yeah good catch, changed

pkg/utils/csv.go
StevenTobin: fixed now

pkg/utils/io.go
b1zzu: @StevenTobin do you think is worth doing this?

StevenTobin: Ah completely missed this. Yeah makes sense, I'll make the change

PR #53, opened 2020-05-08
Add cmd to retrieve the current-csv from the manifests directory

b1zzu:
**What**

The cmd will retrieve the current CSV from the manifests directory containing multiple CSVs

**Verify**

1. Download and extract the manifests directory

```
export SRC_IMAGE='registry-proxy.engineering.redhat.com/rh-osbs/3scale-amp2-3scale-rhel7-operator-metadata@sha256:8355860dbc1052884138612fe1b12a149ceee58440831c041eb3fd45bb38a52e'
mkdir /tmp/mydir
./delorean ews extract-manifests --src-image ${SRC_IMAGE} --extract-dir /tmp/mydir
```
 
2. Retrieve the current CSV from `/tmp/mydir`

```
delorean ews current-csv --directory /tmp/mydir --output /tm/tmpcurrentcsv.json
```

### DIFF
--- a/cmd/currentCSV.go
+++ b/cmd/currentCSV.go
@@ -1,9 +1,7 @@
 package cmd
 
 import (
-	"encoding/json"
 	"fmt"
-	"os"
 
 	"github.com/integr8ly/delorean/pkg/utils"
 	"github.com/spf13/cobra"
@@ -28,26 +26,7 @@ func init() {
 			}
 
 			fmt.Printf("Write current CSV %s to %s\n", file, flags.output)
-			bytes, err := json.Marshal(csv)
-			if err != nil {
-				handleError(err)
-			}
-
-			// truncate the existing file
-			write, err := os.Create(flags.output)
-			if err != nil {
-				handleError(err)
-			}
-
-			_, err = write.Write(bytes)
-			if err != nil {
-				handleError(err)
-			}
-
-			err = write.Close()
-			if err != nil {
-				handleError(err)
-			}
+			utils.WriteObjectToJSON(csv, flags.output)
 		},
 	}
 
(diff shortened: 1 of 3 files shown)


cmd/currentCSV.go
```diff
@@ -0,0 +1,61 @@
...
+		Use:   "current-csv",
+		Short: "Retrieve the current CSV from the manifests directory and write it in JSON format to the output file",
+		Run: func(cmd *cobra.Command, args []string) {
+			csv, file, err := utils.GetCurrentCSV(flags.directory)
+			if err != nil {
+				handleError(err)
+			}
+
+			fmt.Printf("Write current CSV %s to %s\n", file, flags.output)
+			bytes, err := json.Marshal(csv)
```
mikenairn, line 31: Any reason we need to write this as json instead of yaml?

cmd/currentCSV.go
```diff
@@ -0,0 +1,61 @@
...
+
+			_, err = write.Write(bytes)
+			if err != nil {
+				handleError(err)
+			}
+
+			err = write.Close()
+			if err != nil {
+				handleError(err)
+			}
```
mikenairn, line 50: Probably better to add a helper like the one here for the yaml version https://github.com/integr8ly/delorean/blob/master/pkg/utils/io.go#L41

mikenairn:
@b1zzu Does cobra have any support for just returning json/yaml as output from the commands?

cmd/currentCSV.go
b1zzu: It's easy to parse JSON, for example, Jenkins has the built-in function, and cmd like `jq` is available on all distro on opposite to `yq`

But I found out there is a built-in function to read yaml too in Jenkins, so I could also write to yaml

b1zzu:
@mikenairn should be ready now, I've added the util method to write to JSON

PR #54, opened 2020-05-11
feat: custom CIDR ranges, private cluster option

psturc:
### Motivation
https://issues.redhat.com/browse/INTLY-7661

### What was changed
* added [custom OSD network configuration](https://github.com/integr8ly/delorean/compare/master...psturc:ocm-networking-updates?expand=1#diff-30c6bfb85a01f21bab414346c984d3e3R6-R11) (as suggested [here](https://issues.redhat.com/browse/INTLY-7658?focusedCommentId=14088967&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-14088967))
* added option to install private cluster

### How to verify
1. Clone this branch
2. Run `make ocm/cluster.json`
> Printed out cluster configuration should contain the network config [from the template](https://github.com/integr8ly/delorean/compare/master...psturc:ocm-networking-updates?expand=1#diff-30c6bfb85a01f21bab414346c984d3e3R6-R11)
3. Run `make ocm/cluster.json PRIVATE=true`
> You should see `"api": {"listening": "internal"}` in the output (meaning that the cluster with this field will be configured as private)

### DIFF
--- a/templates/ocm/cluster-template.json
+++ b/templates/ocm/cluster-template.json
@@ -3,6 +3,12 @@
     "managed": true,
     "multi_az": false,
     "name": "my_cluster_name",
+    "network": {
+      "machine_cidr": "10.11.128.0/23",
+      "service_cidr": "10.11.0.0/18",
+      "pod_cidr": "10.11.64.0/18",
+      "host_prefix": 23
+    },
     "nodes" : {
       "compute" : 4,
       "compute_machine_type" : {"id" : "m5.xlarge" }
--- a/scripts/ocm/ocm.sh
+++ b/scripts/ocm/ocm.sh
@@ -74,16 +74,22 @@ get_latest_generated_access_key_id() {
 
 create_cluster_configuration_file() {
     local timestamp
+    local listening="external"
 
     : "${OCM_CLUSTER_LIFESPAN:=4}"
     : "${OCM_CLUSTER_NAME:=rhmi-$(date +"%y%m%d-%H%M")}"
     : "${OCM_CLUSTER_REGION:=eu-west-1}"
     : "${BYOC:=false}"
     : "${OPENSHIFT_VERSION:=}"
+    : "${PRIVATE:=false}"
 
     timestamp=$(get_expiration_timestamp "${OCM_CLUSTER_LIFESPAN}")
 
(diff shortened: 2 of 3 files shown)


psturc:
/assign @damienomurchu @tonyxrmdavidson 

This change will allow out pipelines to provision private clusters with required network configuration (as we learned it in https://issues.redhat.com/browse/INTLY-7658)

templates/ocm/cluster-template.json
```diff
@@ -3,6 +3,12 @@
     "managed": true,
     "multi_az": false,
     "name": "my_cluster_name",
+    "network": {
+      "machine_cidr": "10.11.128.0/23",
+      "service_cidr": "10.11.0.0/18",
+      "pod_cidr": "10.11.64.0/18",
+      "host_prefix": 23
+    },
```
davidffrench, line 11: @psturc Do we want to always create clusters from this pipeline using these networking values?

cc @maleck13 @aidenkeating

aidenkeating: @davidffrench To me, this seems fine. They are the most restricted values that we know we have to support. We know `/24` is not possible at the moment

davidffrench: Thanks @aidenkeating

davidffrench:
/lgtm

Changes look good @psturc . I have reviewed code but not verified

psturc:
Thanks for review guys.
I'll merge this PR once the new version of CRO that fixes the issue with CIDR ranges is deployed to staging. Otherwise these changes would break other pipelines that use this tooling.

davidffrench:
@psturc I think this can be merged now

templates/ocm/cluster-template.json
```diff
@@ -3,6 +3,12 @@
     "managed": true,
     "multi_az": false,
     "name": "my_cluster_name",
+    "network": {
+      "machine_cidr": "10.11.128.0/23",
+      "service_cidr": "10.11.0.0/18",
+      "pod_cidr": "10.11.64.0/18",
+      "host_prefix": 23
```
mikenairn, line 10: Are these values always the same? Are they only relevant to a byoc cluster, or any?

psturc: this is the minimal range we're supporting and it works on both byoc and non-byoc

PR #55, opened 2020-05-11
make sure the right branch image tag is used when create quay image tags

wei-lee:


b1zzu:
As this util is growing should we maybe evaluate some semver libraries?
* https://github.com/blang/semver
* https://github.com/Masterminds/semver
* https://github.com/coreos/go-semver

wei-lee:
@b1zzu I did think about and looked at it, but seems a bit overkill for what we need at the moment. But if we have more use case (especially around comparing versions), then we should use a module.

mikenairn:
@wei-lee I actually think we should just make it use the branch name in all situations instead of using major.minor, makes it easier to understand, branch in git maps to tag in quay, same for tags in git.

This chnage doesn't prevent us from making that update after though by updating ReleaseBranchImageTag and updateImageMirroringConfig to use the release branch instead.

Current 2.1 config could be updated manually this time round since it's already been created.

wei-lee:
@mikenairn ok, sounds good to me.

wei-lee:
@mikenairn Can you add "/lgtm" to the PR?

b1zzu:
> @b1zzu I did think about and looked at it, but seems a bit overkill for what we need at the moment. But if we have more use case (especially around comparing versions), then we should use a module.

Yes, I agree but is good to keep it in mind if we start to add also version comparison or other semver futures

PR #57, opened 2020-05-12
Update csv file discovery

mikenairn:
* Allow different naming of csv file (CRW uses csv.yaml)
* Use csv version to determine dest dir (crw is prepending v to the directory)
*  Process current csv during copy (Ensures that all manifests added to a manifest directory are always in the processed order)

### DIFF
--- a/cmd/extractManifests.go
+++ b/cmd/extractManifests.go
@@ -60,6 +60,11 @@ func copyLatestManifest(srcPkgDir, destPkgDir string) (string, string, error) {
 		return "", "", err
 	}
 
+	err = utils.ProcessCurrentCSV(destPkgDir, nil)
+	if err != nil {
+		return "", "", err
+	}
+
 	return srcBundleDir, destBundleDir, nil
 }
 
--- a/cmd/extractManifests.go
+++ b/cmd/extractManifests.go
@@ -48,7 +48,7 @@ func copyLatestManifest(srcPkgDir, destPkgDir string) (string, string, error) {
 	}
 
 	srcBundleDir := filepath.Dir(csvFile)
-	destBundleDir := filepath.Join(destPkgDir, filepath.Base(srcBundleDir))
+	destBundleDir := filepath.Join(destPkgDir, csv.Spec.Version.String())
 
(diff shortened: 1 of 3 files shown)


b1zzu:
:eyes:

cmd/extractManifests.go
```diff
@@ -60,6 +60,11 @@ func copyLatestManifest(srcPkgDir, destPkgDir string) (string, string, error) {
 		return "", "", err
 	}
 
+	err = utils.ProcessCurrentCSV(destPkgDir, nil)
```
b1zzu, line 63: @mikenairn I don't get what does this change, is it just removing the "status" and "creationTimestamp" from the current csv?

mikenairn: It should hopefully change nothing. Problem is since changing to update the csv by marshalling/unmarshelling a GO object, the ordering of the CSV is always changed, seems to be no way round it. This just ensures that anything we do always produces the ordered version of the CSV in the destination manifest directory.

tbh, i really don't like that the csv get changed so much, I'd rather we didn't change them at all. Also noticed that some things are getting removed (relatedImages), i presume because we are using a different CSV struct from what created it, which is less than ideal.

b1zzu: Ok, I see, yes I had the same thought when I was developing the osdAddonRelease cmd.

I found out that with yaml it should be possible to use the object `yaml.MapSlice` which should preserve the order of the keys, we could then built around some helper to edit specific keys, similar to the yq format, without therefore losing unknown keys or shuffling the yaml file, we could do some experiment separately

mikenairn: Yes, i think we need to come back to this. I thought it was just ordering, which i could just about get over even though it wasn't ideal, but now that i know things could potentially be removed i'm a bit more concerned.

b1zzu: Yes, you never know if the thing you have removed was a critical field, but this is not only this PR problem, everywhere we are parsing, edit and writing yaml files could have the same problem

mikenairn: https://issues.redhat.com/browse/DEL-278

PR #58, opened 2020-05-12
Del 242 test containers

wei-lee:
To verify:

1. Make sure a RHMI cluster is available and running, and you have access to the `kubeconfig` file and it is saved locally somewhere.
2. Check out the branch and build it. Run the following command:
    ```
    export KUBECONFIG=<path to kubeconfig file>
     ./delorean tests run --output /tmp/tests --test-config ./test-containers.yaml
    ```
3. Wait for it to complete. It will take about 15m for the tests containers to complete. When they are finished, you should be able to access the results and logs in the output directory.

Review this commit: 0e1169029be09aa9398c36834febd871700443b1


b1zzu:
:eyes:

test-containers.yaml
```diff
@@ -0,0 +1,14 @@
+---
+tests:
+  - name: codeready-test
+    image: quay.io/crw/osd-e2e
+    timeout: 3600
+    envVars:
+      - name: CODEREADY_NAMESPACE
```
b1zzu, line 7: Should we try to unify the namespace? If we always use `NAMESPACE` we are limited to one for test container, if we prefix the namespace with the product name `CODEREADY_NAMESPACE`, `THREESCALE_NAMESPACE`, we could pass all namespaces to all test containers, and the container will decide which one to use, wdyt?

cmd/tests.go
```diff
@@ -0,0 +1,283 @@
...
+	if err != nil {
+		return nil, err
+	}
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return nil, err
+	}
+
+	dt := time.Now()
+	outputDir := path.Join(f.outputDir, dt.Format("2006-01-02-03-04-05"))
```
b1zzu, line 98: I'm not sure that we should create a subdirectory with an autogenerated name because if we run the tests from a pipeline and then we want to archive the test results, we will have to archive also the subdirectory or to implement an additional step to deal with it. And generally, I don't think we will run these tests multiple times from the same machine, maybe we should just have a step checking if the output-dir is empty or doesn't exist (means we create it) befor starting the tests, wdyt?

cmd/tests.go
```diff
@@ -0,0 +1,283 @@
...
+	fmt.Println("[Prepare] Create ClusterRoleBinding for the service account")
+	gvk := schema.FromAPIVersionAndKind("v1", "namespace")
+	owner := metav1.NewControllerRef(ns, gvk)
+	if _, err = utils.CreateClusterRoleBinding(c.clientset, sa, "cluster-admin", *owner); err != nil {
+		return err
+	}
+	var wg sync.WaitGroup
+	for _, testContainer := range c.tests {
+		wg.Add(1)
+		go func(t TestContainer) {
```
b1zzu, line 138: Is this running all tests in parallel? Are we sure is a good idea, because if we want to add also the integreatly tests to it we will for sure have some conflict (this if we don't plan to create a separate yaml for the tests that should run in sequence)

cmd/tests.go
```diff
@@ -0,0 +1,283 @@
...
+func (c *runTestsCmd) runTestContainer(ctx context.Context, test TestContainer) (bool, error) {
+	job := getTestContainerJob(c.namespace, test)
+	if _, err := utils.CreateJob(c.clientset, job); err != nil {
+		return false, err
+	}
+	podSelector := fmt.Sprintf("job-name=%s", job.GetName())
+	var podList *v1.PodList
+	var err error
+	fmt.Println(fmt.Sprintf("[%s] Waiting for job to be started", test.Name))
+	err = wait.PollImmediate(time.Duration(1)*time.Second, time.Duration(60)*time.Second, func() (done bool, err error) {
```
b1zzu, line 170: Should we deprecate the `wait.Retry` function in our pkg/utils in favor of the k8s method?

cmd/tests.go
```diff
@@ -0,0 +1,283 @@
...
+	to := path.Join(c.outputDir, testName)
+	if err := os.MkdirAll(to, os.ModePerm); err != nil {
+		return err
+	}
+	oc := &utils.OCRunner{Kubeconfig: c.kubeconfig}
+	return oc.Run("rsync", from, to, "-c", "sidecar", "-n", c.namespace)
+}
+
+func (c *runTestsCmd) downloadLogs(pod v1.Pod, testName string) error {
+	outputFile := path.Join(c.outputDir, testName, "logs", "container.log")
```
b1zzu, line 272: I would be more for putting the `test-run-results/` in a subdirectory instead of the `container.log` or just put both in the same one

cmd/tests.go
```diff
@@ -0,0 +1,283 @@
...
+}
+
+func (c *runTestsCmd) downloadTestResults(pod v1.Pod, testName string) error {
+	from := fmt.Sprintf("%s:/test-run-results/", pod.GetName())
+	to := path.Join(c.outputDir, testName)
+	if err := os.MkdirAll(to, os.ModePerm); err != nil {
+		return err
+	}
+	oc := &utils.OCRunner{Kubeconfig: c.kubeconfig}
+	return oc.Run("rsync", from, to, "-c", "sidecar", "-n", c.namespace)
```
b1zzu, line 268: I was hoping we could have used the openshift api without calling the`oc` cmd, but if it works is fine

b1zzu:
When I run the cmd I see two warnings:
```
Using config file: /home/b1zzu/.delorean.yaml
[Prepare] Create namespace rhmi-product-tests
[Prepare] Create serviceAccount cluster-admin-sa
[Prepare] Create ClusterRoleBinding for the service account
[3scale-test] Start test container
[codeready-test] Start test container
[3scale-test] Waiting for job to be started
[codeready-test] Waiting for job to be started
[codeready-test] Pod found for job: codeready-test-mrnkr
[codeready-test] Wait for test container to finish
[3scale-test] Pod found for job: 3scale-test-2v5xg
[3scale-test] Wait for test container to finish
[3scale-test] Pod=3scale-test-2v5xg State=Pending
[codeready-test] Pod=codeready-test-mrnkr State=Pending
[3scale-test] Pod=3scale-test-2v5xg State=Pending
[codeready-test] Pod=codeready-test-mrnkr State=Pending
[3scale-test] Pod=3scale-test-2v5xg State=Pending
[codeready-test] Pod=codeready-test-mrnkr State=Pending
[codeready-test] Pod=codeready-test-mrnkr State=Running
[codeready-test] Pod=codeready-test-mrnkr Container=test ContainerStatus=Running
[3scale-test] Pod=3scale-test-2v5xg State=Running
[3scale-test] Pod=3scale-test-2v5xg Container=test ContainerStatus=Running
[codeready-test] Pod=codeready-test-mrnkr State=Running
[codeready-test] Pod=codeready-test-mrnkr Container=test ContainerStatus=Terminated
[codeready-test] Tests completed. Exit code = 1
[codeready-test] Download test results
WARNING: rsync command not found in path. Please use your package manager to install it.
./codeready-addon-logs.log
./debug_tests.json
./codeready_postgres-67d65db79f-2kfxx.log
./codeready_keycloak-559458bfcf-gvs56.log
./codeready_devfile-registry-58cbd6787f-cg9gj.log
./codeready_plugin-registry-6974f58d59-4649c.log
./codeready_codeready-6ffd8bf4ff-tq55r.log
./junit-che-operator.xml
./addon-metadata.json
[codeready-test] Download test container logs
[codeready-test] Delete test job
[codeready-test] Test container failed
[3scale-test] Pod=3scale-test-2v5xg State=Running
[3scale-test] Pod=3scale-test-2v5xg Container=test ContainerStatus=Terminated
[3scale-test] Tests completed. Exit code = 2
[3scale-test] Download test results
WARNING: rsync command not found in path. Please use your package manager to install it.
./junit.xml
[3scale-test] Download test container logs
[3scale-test] Delete test job
[3scale-test] Test container failed
[Reporting] Tests completed. Results can be found in /tmp/tests/2020-05-13-04-34-46
[TearDown] Delete namespace rhmi-product-tests
```
but they logs where still synced, so I don't know why is printing the WARNING

test-containers.yaml
wei-lee: I'd prefer to just use `NAMESPACE`. Why the tests for one product need to know namespaces of others?

Either way, this can be easily changed in the config file, so we don't need to update implementation.

cmd/tests.go
wei-lee: The main reason behind it is to be able to run the tests outside pipelines as well. In this case, you don't end up the test results override each other.

I don't understand why it is a problem to run inside pipelines? You should archive the `outputDir` any, that will include all subdirs automatically.

cmd/tests.go
wei-lee: Yes, they will running in parallel. I think this is fine for at least products testings. If for some reason the tests can't be run parallel, then we should ask why. Because in production environments, customers will use multiple  products at the same time.

The integreatly tests should be run separately. Either as a different step in the pipeline, or have a separate config file.

cmd/tests.go
wei-lee: Not sure I understand what you mean here. This is for container logs and they are in a different directory?

cmd/tests.go
wei-lee: unfortunately there's no such apis.

cmd/tests.go
wei-lee: sure

wei-lee:
@b1zzu @mikenairn this PR is ready now.

test-containers.yaml
b1zzu: If we want I can talk with crw guys to change the ENV name, at the end they have created because of us, so they shouldn't have any problem in changing it

cmd/tests.go
b1zzu: I just think that it will be better if the `test-run-results` are in a subdirectory instead of the `container.log`, like this:
```
outdir/
    3scale/
         results/
             junit.xml
         container.log
    crw/
         results/
              junit.xml
              other.log
         container.log
```
this mainly if we want to avoid the container.log to accidentally overwrite the logs produced by the test container

cmd/tests.go
b1zzu: Yeh, I agree is good to make them run in parallel, to simulate a real case

cmd/tests.go
b1zzu: Just wanted to reduce that additional incapsulation (sub-directory), when running in a pipeline, but fine like this

test-containers.yaml
wei-lee: The way I see it each team will be responsible for maintaining the section of the configuration file. It's up to them to decide what namespace they want to use. I don't know if we need to care that much. Also why do we ask them to add this env when their tests don't need it?

cmd/tests.go
wei-lee: fine. I will create separate dirs for results and logs then.

cmd/tests.go
merlante: Running tests in series seems like a good idea if,
a) it's possible for the command to fail because of something that happens in a test. At least we can say "it always fails somewhere in the middle of running the CRW tests", for example, if the images are run in order.
b) a lack of determinism in the running order of tests could make it harder to reproduce a failure.

Cross product testing, in another phase of the pipeline, seems like the logical place to test multiple products at the same time.

b1zzu:
@wei-lee  for me is ready when you want you can merge it

/lgtm

cmd/tests.go
b1zzu: @merlante are you suggesting we should have a way to run them sequentially if they fails and we have to debug the failure in a deterministic way?

wei-lee:
/approve

thanks @b1zzu

cmd/tests.go
wei-lee: @merlante we could add support for running the test in sequence, but I am not convinced that it is the right way to run these tests. For each of the tests, they are quite self-contained to run only against the product, and if these tests have a dependency on another product, I would see that as an issue. Also running them in sequence will dramatically increase the time to run our pipelines and it will get longer and longer if more product tests are added.

PR #60, opened 2020-05-13
Approve RHMI installplan

psturc:
### Motivation
Since there's currently required to approve the RHMI installplan so the installation can continue, I've added an `oc` command to do the approval once the installplan is available.

Even if the approval would be set back to "automatic" in the future, this command won't fail (it will finish with 0 code and a message `installplan.operators.coreos.com/install-xxxxx patched (no change)`)

### DIFF
--- a/scripts/ocm/ocm.sh
+++ b/scripts/ocm/ocm.sh
@@ -126,13 +126,18 @@ install_rhmi() {
     local rhmi_name
     local infra_id
     local csv_name
+    local installplan_name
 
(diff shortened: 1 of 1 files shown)


davidffrench:
@psturc The MR is merged now to ensure this is auto approved on the addon flow again - https://gitlab.cee.redhat.com/service/managed-tenants/merge_requests/145. 

Which means this change is not required any longer. I would prefer we didn't add it to be honest so if there is some regression with the add on flow, the pipeline will pick it up.

PR #67, opened 2020-05-20
update the osd-addon command to work with the new MT repo structure

wei-lee:
https://issues.redhat.com/browse/DEL-284

Here are some example MRs created by using the updated cli: 
* stage - https://gitlab.cee.redhat.com/integreatly-qe/managed-tenants/merge_requests/13
* edge - https://gitlab.cee.redhat.com/integreatly-qe/managed-tenants/merge_requests/14


cmd/osdAddonRelease.go
```diff
@@ -115,6 +116,24 @@ type osdAddonReleaseCmd struct {
 	gitPushService         services.GitPushService
 }
 
+type addon struct {
+	content string
+}
+
+func (a *addon) SetCurrentCSV(currentCSV string) {
```
b1zzu, line 123: Can we keep this method private too?

cmd/osdAddonRelease.go
```diff
@@ -115,6 +116,24 @@ type osdAddonReleaseCmd struct {
...
+	r := regexp.MustCompile(`currentCSV: integreatly-operator\..*`)
+	s := r.ReplaceAllString(a.content, fmt.Sprintf("currentCSV: %s", currentCSV))
```
b1zzu, line 125: Couldn't we use the unstructured yaml instead of doing a regex replacement?

b1zzu:
@wei-lee For me is good, we can merge it to unlock the 2.2.0 release, and we could solve my comments in a separate PR (if we should do something for them)

cmd/osdAddonRelease.go
wei-lee: yes, we should.

cmd/osdAddonRelease.go
wei-lee: I did think about it, but at the end I think using the regex here is the best option for a few reasons:

1. If we load the whole file and unmarshal it into an object, we will not be able to control the order of the objects in the output file, or the format of certain objects. We could end up with unnecessary changes to this file.
2. All we really change is just this one line.

So for now using the regex is the safest and simplest solution. We could it in the future if we have to.

wei-lee:
@b1zzu if you can add another `/lgtm` then this is ready for merging.

cmd/osdAddonRelease.go
b1zzu: Maybe we should look into extending the unstructured object to support preserving the yaml order

wei-lee: if you are talking about the k8s's unstructured object, that doesn't work for us as the object here is not a k8s's API object. The unstructured object only works for k8s API objects. We would have to just use `map[string]interface{}` type to load it.

PR #70, opened 2020-05-21
rename csv file to csv.j2 file and only push the new branch

wei-lee:


cmd/osdAddonRelease.go
```diff
@@ -480,3 +490,12 @@ func (c *osdAddonReleaseCmd) udpateTheCSVManifest(channel releaseChannel) (strin
 	}
 	return relative, nil
 }
+
+func (c *osdAddonReleaseCmd) renameCSVFile(channel releaseChannel) (string, error) {
+	o := fmt.Sprintf("%s/%s/%s.v%s.clusterserviceversion.yaml", channel.bundlesDirectory(), c.version.Base(), "integreatly-operator", c.version.Base())
+	n := fmt.Sprintf("%s/%s/%s.v%s.clusterserviceversion.yaml.j2", channel.bundlesDirectory(), c.version.Base(), "integreatly-operator", c.version.Base())
```
mikenairn, line 496: What's the reason for this? Is this something to do with the new changes being made in the managed tenants repo?

mikenairn:
/lgtm

Just a question about why we need to do it.

cmd/osdAddonRelease.go
wei-lee: yes, that's right. The MT repo is now using a template: https://gitlab.cee.redhat.com/service/managed-tenants/blob/master/addons/integreatly-operator/bundles/2.1.1/integreatly-operator.v2.1.1.clusterserviceversion.yaml.j2 and it will be responsible for generating the real CSV file.

PR #77, opened 2020-05-28
fix: remove access keys rotation

psturc:
### Motivation
https://issues.redhat.com/browse/INTLY-7975

We no longer should rotate credentials for osdCcsAdmin user.
That means it is not necessary to have the functionality for generating access keys in ocm

### What was done in this PR
* removed functionality for generating AWS access key for `osdCcsAdmin` (`create_access_key` function)
* using exported variables with AWS credentials for generating cluster configuration instead from the file created by `create_access_key` function
* updated docs

### How to verify
1. Get some AWS account (!currently unused!) and create and export AWS access key & secret for `osdCcsAdmin` user as described in the [readme](https://github.com/integr8ly/delorean/pull/77/files#diff-a2e6eb7ae52f8791ee3b37cbd4fa0d38)
2. Also make sure `BYOC=true` is exported
3. Run `make ocm/cluster.json` and verify that the part of the config containing the AWS credentials contains the same values as those you exported

### DIFF
--- a/docs/ocm/README.md
+++ b/docs/ocm/README.md
@@ -2,16 +2,15 @@
 
 If you want to test your changes on a cluster, the easiest solution would be to spin up OSD 4 cluster using `ocm`. If you want to spin up a cluster using BYOC (your own AWS credentials), follow the additional steps marked as **BYOC**.
 
-#### Prerequisites
+### Prerequisites
 * [OCM CLI](https://github.com/openshift-online/ocm-cli/releases)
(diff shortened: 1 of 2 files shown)


pawelpaszki:
looks reasonable to me

pawelpaszki:
I will let others take a look as well
/lgtm

psturc:
Thanks for review @pawelpaszki. I would like to get this merged before the nightly pipelines are triggered again, i.e. in the evening, if nobody is against it.

psturc:
@austincunningham @aidenkeating fyi these are the changes to the tooling we are using in pipelines that I mentioned at the standup.

PR #78, opened 2020-05-28
DEL-286 Process Images for GA and non Ga cases

laurafitzgerald:
**JIRA**
https://issues.redhat.com/browse/DEL-286

Note: Pushing the code changes for verification as I've change the flow somewhat. Tests to follow.

**What**
update process-csv-images command to allow for processing of images in ga and non ga cases

**How**
I did a refactor here so that we can reuse the GetAndUpdateOperandImages
and the GetAndUpdateOperatorImages
In both cases the flow is based on whether it's Ga or not

_If it's GA_ 
Check for any stage images and process those to prod images

_It it's NotGA_ 
check for any stage or prod images and process those to delorean(aka quay) images

**Verification**

_**Preparation Steps**_
Build the cli
`make build/cli`
Extract Manifest from an image
`./delorean ews extract-manifests --src-image registry-proxy.engineering.redhat.com/rh-osbs/3scale-amp2-3scale-rhel7-operator-metadata:1.11.0-82 --extract-dir $EXTRACT_DIR`
Extract the Manifest
`./delorean ews extract-manifests --src-dir $EXTRACT_DIR --dest-dir $MANIFEST_DIR`

At this stage go to the integreatly-operator dir that you are using and do a git add manifests/ so that you can compare the chagnes below.

**_Case GA:_**
- Run Preparation steps
- Run the command below
`./delorean ews process-csv-images --manifest-dir $THREESCALE_MANIFEST_DIR --isGa=True`
- Check diff
`git diff manifests/`

Verify that any stage images for container envs, annotation and operator image are translated to prod images (aka have `.stage` removed)

**_Case NotGA_**
- Run Preparation steps
- Run the command below
`./delorean ews process-csv-images --manifest-dir $THREESCALE_MANIFEST_DIR --isGa=False`
- Check diff
`git diff manifests/`

Verify that any stage images for container envs, annotation and operator image are translated to delorean (aka have a quay image reference )


pkg/utils/csv.go
```diff
@@ -346,32 +346,32 @@ func GetAndUpdateOperandImagesToDeloreanImages(manifestDir string, extraImages [
...
-			//found a registry.stage.redhat.io image
-			if stageMatched != "" {
-				matched = stageMatched
-			}
+			//if it's GA, check for stage images and replace them.
+			if isGa {
+				stageMatched := ReStage.FindString(env.Value)
+				if stageMatched != "" {
+					prodImage := env.Value
+					prodImage = strings.Replace(prodImage, ".stage", "", 1)
```
wei-lee, line 354: it's best to replace using the full regex. While they are doing the same thing, using the full regex makes the code a little bit easier to understand. Also if this url ever needs to be changed, we only need to modify it once.

laurafitzgerald: Ah actually, I created a new method for that and mean to push this down to that so it would only need to be edited once.

pkg/utils/csv.go
```diff
@@ -346,32 +346,31 @@ func GetAndUpdateOperandImagesToDeloreanImages(manifestDir string, extraImages [
...
+				deloreanMatched := ReDelorean.FindString(env.Value)
+				if deloreanMatched != "" {
+					continue
+				}
+				prodMatched := ReProd.FindString(env.Value)
+				stageMatched := ReStage.FindString(env.Value)
+				//found a registry.redhat.io image
+				if prodMatched != "" || stageMatched != "" {
+					deloreanImage := BuildDeloreanImage(env.Value)
+					mirrorString := BuildOSBSImage(env.Value) + " " + deloreanImage
```
laurafitzgerald, line 368: This is used in a couple of places too. Could also be moved out to it's method.

pkg/utils/testdata/validManifests/3scale3/0.5.2/apps_v1alpha1_apimanager_crd.yaml
```diff
@@ -0,0 +1,238 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
mikenairn, line 1: I doubt we need to add all the crd files, we only manipulate the csvs , so should be enough just to include that for test data.

mikenairn:
Tested it locally, seems to be working as required.
I removed a test that wasn't working properly (Would only work the first time you ran it), removed the extra testdata files, and added a couple of additional tests to check the images existed in the CSV as expected (GA or not GA).

/lgtm
/approve

PR #79, opened 2020-05-29
Add AMQ test container

b1zzu:
I already tried it manually and it's working and is a simple smoke test, so I think we should add it to our tests

### DIFF
--- a/test-containers.yaml
+++ b/test-containers.yaml
@@ -13,7 +13,7 @@ tests:
       - name: NAMESPACE
         value: redhat-rhmi-3scale
   - name: amq-test
-    image: quay.io/enmasse/enmasse-test-container
+    image: quay.io/enmasse/enmasse-test-container:0.31.2
     timeout: 3600
     envVars:
       - name: KUBERNETES_NAMESPACE
--- a/test-containers.yaml
+++ b/test-containers.yaml
@@ -12,3 +12,9 @@ tests:
     envVars:
       - name: NAMESPACE
         value: redhat-rhmi-3scale
+  - name: amq-test
+    image: quay.io/enmasse/enmasse-test-container
+    timeout: 3600
+    envVars:
+      - name: KUBERNETES_NAMESPACE
+        value: redhat-rhmi-amq-online


test-containers.yaml
```diff
@@ -11,4 +11,10 @@ tests:
     timeout: 3600
     envVars:
       - name: NAMESPACE
-        value: redhat-rhmi-3scale
\ No newline at end of file
+        value: redhat-rhmi-3scale
+  - name: amq-test
+    image: quay.io/enmasse/enmasse-test-container
```
mikenairn, line 16: Are they just pushing to latest all the time or are they intending to version it?

mikenairn: I just realised this is on delorean, why is this file in here? @b1zzu @wei-lee

b1zzu: I have created the PR also against the operator: https://github.com/integr8ly/integreatly-operator/pull/822, and I don't know if we should keep just one to avoid confusion

b1zzu: > Are they just pushing to latest all the time or are they intending to version it?
David should tag the image, then I would update the PR with the correct tag to avoid running always the latest version

mikenairn: I don't see why we need two files, or one is in this repo at all, the one in the operator should be the one being used for testing.

I'd be in favour of removing this one from here unless I'm missing the point of why it was added here in the first place.

b1zzu: We could keep one in the delorean repo as an example but the "master" one should always be the one in the integreatly-operator

but if this creates confusion then we should just keep the one in the integreatly-operator

mikenairn: OK, i just don't understand why it's here at all tbh. We only want to be updating the one in the intly operator repo with updates anyway.

b1zzu:
Because the `test-containers.yaml` in the Delorean project exists only as an example, it doesn't make sense to update it which each new test container

test-containers.yaml
b1zzu: Let's close this PR and focus on the one to the operator: https://github.com/integr8ly/integreatly-operator/pull/822

PR #81, opened 2020-05-29
Add rsync to Dockerfile

mikenairn:
* Add rsync to Dockerfile
* Update e2e-test-extract-creds script to wait on setup container completion


laurafitzgerald:
@mikenairn the change looks good. Should we try to build it to test it? I had some problem with the COPY of the crt but not sure if it's expected that this be built somewhere else?

mikenairn:
> Should we try to build it to test it? I had some problem with the COPY of the crt but not sure if it's expected that this be built somewhere else?

@laurafitzgerald You can use this make target to build and test the image `make image/test`. I'd guess you are running it form the wrong directory.

PR #82, opened 2020-06-01
DEL-296 Update the zz_generated_metadata.branch value when creating the ci config file for the release branch

b1zzu:
**What**
Update the `zz_generated_metadata.branch` value when creating the OpenShift CI config file for the release branch
Ref: https://github.com/openshift/release/blob/master/ci-operator/config/integr8ly/integreatly-operator/integr8ly-integreatly-operator-master.yaml#L68

**Plus**: Added the `UnstructYaml` object to modify a yaml file without changing the order of the fields or removing any fields

### DIFF
--- a/cmd/openshiftCIRelease.go
+++ b/cmd/openshiftCIRelease.go
@@ -4,6 +4,13 @@ import (
 	"bufio"
 	"context"
 	"fmt"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path"
+	"strings"
+	"time"
+
 	"github.com/go-git/go-git/v5"
 	"github.com/go-git/go-git/v5/config"
 	"github.com/go-git/go-git/v5/plumbing"
@@ -13,12 +20,6 @@ import (
 	"github.com/integr8ly/delorean/pkg/services"
 	"github.com/integr8ly/delorean/pkg/utils"
 	"github.com/spf13/cobra"
-	"io/ioutil"
-	"os"
-	"os/exec"
-	"path"
-	"strings"
-	"time"
 )
 
 const (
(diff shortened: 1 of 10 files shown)


pkg/utils/io.go
```diff
@@ -135,21 +117,5 @@ func FileExists(filename string) bool {
...
-		}
-	}
-	err = f.Close()
-	if err != nil {
-		return err
-	}
-
-	return nil
+	return ioutil.WriteFile(writePath, []byte(strings.Join(content, "\n")), 0644)
 }
```
mikenairn, line 121: Were any of these changes required? If all these utility functions are just calling another utility function, is there any point in keeping them?

cmd/openshiftCIRelease.go
```diff
@@ -249,16 +249,22 @@ func updateCIOperatorConfig(repoDir string, version *utils.RHMIVersion) error {
...
+		return err
+	}
+
+	err = y.Set("zz_generated_metadata.branch", version.ReleaseBranchName())
+	if err != nil {
+		return err
+	}
 
-	err = ioutil.WriteFile(releaseConfig, []byte(releaseConfigContents), 0644)
+	err = y.Write(releaseConfig)
```
mikenairn, line 267: Seems handy, can probably be used elsewhere. I assume we don't end up wildly different config files when doing this?

pkg/utils/io.go
b1zzu: Not really, the only thing this function is doing is joining the array `strings.Join(content, "\n")`, I can remove it and also fix the place where is called it, I don't mind either solution

cmd/openshiftCIRelease.go
b1zzu: yes, the idea is that we should be able to reuse it in any places where we are editing yaml files, but I would like to avoid stack too many changes in one single PR, I can create a task to do this refactoring

mikenairn:
/lgtm

You could consider squashing the commits before merge, but changes look fine.

PR #89, opened 2020-06-05
added jq to delorean image

pawelpaszki:
## What
added jq to delorean image


mikenairn:
Looks fine, but can you add a check for jq here https://github.com/integr8ly/delorean/blob/master/scripts/image/test#L22

pawelpaszki:
@mikenairn - done

PR #92, opened 2020-06-11
Del 300 product downtime

wei-lee:
https://issues.redhat.com/browse/DEL-300

to verify:

1. Prepare  a RHMI cluster and have the workload app installed.
2. Run the following command from this branch:
   ```
   ./delorean pipeline query-report \
      --config-file ./configurations/downtime-report-config.yaml \
      -o /tmp
   ```
3. Once finished, check the output file


b1zzu:
:eyes:

cmd/productTests.go
```diff
@@ -58,8 +58,8 @@ type runTestsCmd struct {
 func init() {
 	f := &runTestsCmdFlags{}
 	cmd := &cobra.Command{
-		Use:   "run",
-		Short: "Execute test containers",
+		Use:   "product-tests",
```
b1zzu, line 61: We should look where the `delorean tests run` is documented and update it

b1zzu: we should fix the Makefile in the integreatly-operator otherwise the nightly pipelines will fails

b1zzu:
/lgtm

But we should fix or at least have a PR for the integreatly-operator before merging this one

cmd/productTests.go
wei-lee: yes, I will do that

wei-lee:
@b1zzu need to rebase so if you can approve again that will be great. Also I linked it to operator PR to update the make target too. Thanks.

STEPHINRACHEL:
Verified locally 
The output file generated can be found [here](https://gist.github.com/STEPHINRACHEL/b72dac26143939f3c0e424a627427fa3)
/approve

PR #97, opened 2020-06-20
INTLY-6593 added delorean pipeline export-results command

StanleyKaleta:


cmd/exportResults.go
```diff
@@ -0,0 +1,73 @@
...
+	jobLink			string
+	accessKeyId		string
+	secretAccessKey string
+}
+
+func init() {
+	f := &testResultsCmdFlags{}
+	cmd := &cobra.Command{
+		Use:   "export-results",
+		Short: "Export RHMI product test results to s3 bucket",
```
b1zzu, line 29:
```suggestion
		Short: "Export RHMI and product test results to s3 bucket",
```
we should try to explain that it should export also the RHMI tests and not only the product tests

cmd/exportResults.go
```diff
@@ -0,0 +1,73 @@
...
+
+			// generate metadata json inside results dir
+			metadataFile := f.testResultsDir + "metadata.json"
+			fmt.Println("Generating... " + metadataFile)
+
+			err := utils.WriteObjectToJSON(MetaData{
+				Name:    f.pipelineName,
+				Version: f.rhmiVersion,
+				Job: f.jobLink,
+			}, metadataFile)
```
wei-lee, line 40: I am wondering should the metadata.json file generated by a pipeline (e.g. via groovy script)? It will be much easier to modify the content of that file. @b1zzu wdyt?

cmd/exportResults.go
```diff
@@ -0,0 +1,73 @@
...
+			if err != nil {
+				handleError(err)
+			}
+
+			// zip the results
+			zipFile := f.pipelineName + "-" +
+				time.Now().Format("2006-01-02-03-04-05") +
```
b1zzu, line 47: Just to be sure a file with the same name doesn't already exist we could add a four-digit random string
```
b := make([]byte, n)
rand.Read(b)
fmt.Println(string(b))
``` 
wdyt?

pkg/utils/s3export.go
```diff
@@ -0,0 +1,62 @@
...
+
+func AddFileToS3(dir string, fileName string, accessKeyId string, secretAccessKey string) error {
+
+	s, err := session.NewSession(&aws.Config{
+		Region: aws.String(S3_REGION),
+		Credentials: credentials.NewStaticCredentials(
+			accessKeyId,
+			secretAccessKey,
+			""),
+	})
```
wei-lee, line 27: use `session.Must(session.NewSession())`. This will allow configure the AWS client via standard AWS env vars.

pkg/utils/s3export.go
```diff
@@ -0,0 +1,62 @@
...
+	if err != nil {
+		return err
+	}
+	defer file.Close()
+
+	// Get file size and read the file content into a buffer
+	fileInfo, _ := file.Stat()
+	var size int64 = fileInfo.Size()
+	buffer := make([]byte, size)
+	file.Read(buffer)
```
wei-lee, line 49: you can just use `ioutils.ReadAll`

pkg/utils/s3export.go
```diff
@@ -0,0 +1,62 @@
...
+
+	// Config settings of the file.
+	_, err = s3.New(s).PutObject(&s3.PutObjectInput{
+		Bucket:               aws.String(S3_BUCKET),
+		Key:                  aws.String(fileName),
+		ACL:                  aws.String("private"),
+		Body:                 bytes.NewReader(buffer),
+		ContentLength:        aws.Int64(size),
+		ContentType:          aws.String(http.DetectContentType(buffer)),
+		ContentDisposition:   aws.String("attachment"),
```
wei-lee, line 59: you don't need to provide all the fields. YOu just need to provide the fields that are marked `required`

pkg/utils/s3export.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/s3"
+)
+
+const (
+	S3_REGION = "eu-west-1"
+	S3_BUCKET = "rhmi-test-data-storage-tf56"
+)
+
+func AddFileToS3(dir string, fileName string, accessKeyId string, secretAccessKey string) error {
```
wei-lee, line 19: this function is not really unit-testable

pkg/utils/zip.go
```diff
@@ -0,0 +1,64 @@
...
+
+	outFile, err := os.Create(baseFolder + zipFileName)
+	if err != nil {
+		return err
+	}
+	defer outFile.Close()
+
+	w := zip.NewWriter(outFile)
+
+	err = addFiles(w, baseFolder, "")
```
wei-lee, line 20: you are writing the output zip file to the zip file again.

pkg/utils/s3export.go
b1zzu, line 19: Because we will have also a `ReadFile` and probably a `FindFile` or `ListFiles` functions, I think we should create an AWS S3 client structure 

```golang
s3 := NewS3Client(accessKeyId string, secretAccessKey string, bucket string, region string)
s3.AddFile(...)
```
wdyt?

pkg/utils/zip.go
```diff
@@ -0,0 +1,64 @@
...
+import (
+	"archive/zip"
+	"io/ioutil"
+	"os"
+	"strings"
+)
+
+func ZipWriter(baseFolder string, zipFileName string) error {
+
+	outFile, err := os.Create(baseFolder + zipFileName)
```
wei-lee, line 12: it's better to ask for the absolute path of the output zip file to make this function reusable.

pkg/utils/zip.go
```diff
@@ -0,0 +1,64 @@
...
+	err = addFiles(w, baseFolder, "")
+	if err != nil {
+		return err
+	}
+
+	err = w.Close()
+	return err
+}
+
+func addFiles(w *zip.Writer, basePath, baseInZip string) error {
```
wei-lee, line 29: need to valid `basePath` is a directory

pkg/utils/s3export.go
b1zzu: This should also make it easier to mock it for testing

pkg/utils/zip.go
```diff
@@ -0,0 +1,64 @@
...
+	if err != nil {
+		return err
+	}
+
+	for _, file := range files {
+		if !file.IsDir() {
+			// skipping .zip file to be generated
+			if strings.Contains(file.Name(), ".zip") {
+				continue
+			}
```
wei-lee, line 41: I think I know why you are doing this, but I don't think this is a good idea. It is completely normal for a zip file to container another zip file.

wei-lee:
Also try to separate the changes to add dependencies (update the vendor dir) to another commit. This will make it easier for others to focus on only review what's required.

cmd/exportResults.go
StanleyKaleta: that's why I also added seconds there, don't think there's chance that 2 jobs from same pipeline will ever want to report in same second

pkg/utils/s3export.go
StanleyKaleta: ok

pkg/utils/s3export.go
StanleyKaleta: @wei-lee that's what I wanted to avoid, because pipeline can use different aws vars, but we want to use same for reporting

cmd/exportResults.go
b1zzu: My idea was to let the CLI generating it so that we can have a shard structure and if we change the metadata structure we can easily update it everwhere is used, so at least for now that we have just tree values I would keep it like this, wdyt?

pkg/utils/s3export.go
b1zzu: We should use `viper` in the same way we did in all other cmd

cmd/exportResults.go
b1zzu: Ok, it should be enough then :)

pkg/utils/s3export.go
b1zzu: Try to align with @wei-lee PR #98

pkg/utils/s3export.go
```diff
@@ -0,0 +1,62 @@
...
+	"os"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/s3"
+)
+
+const (
+	S3_REGION = "eu-west-1"
+	S3_BUCKET = "rhmi-test-data-storage-tf56"
```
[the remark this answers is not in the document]
b1zzu: even better we should make the Bucket configurable using cmd flags (I don't think we need the region)

cmd/exportResults.go
wei-lee: ok

pkg/utils/s3export.go
StanleyKaleta: @b1zzu what do you want use viper for?

StanleyKaleta: yes, will look at it now

pkg/utils/zip.go
StanleyKaleta: @wei-lee I'd like to create zip file inside the folder which is zipped, so I added this check to skip the zip file

wei-lee: @StanleyKaleta why this comment is resolved without any actions?

StanleyKaleta: @wei-lee it is updated to only omit zip file with the same name as that being created

pkg/utils/zip.go
StanleyKaleta: @wei-lee I think it is validated in the following line 		
`files, err := ioutil.ReadDir(basePath)`

pkg/utils/aws.go
```diff
@@ -32,3 +37,40 @@ func DownloadS3ObjectToTempDir(ctx context.Context, downloader s3manageriface.Do
 	}
 	return o, nil
 }
+
+func CreateSession(accessKeyId string, secretAccessKey string, s3Region string) (*session.Session, error) {
```
wei-lee, line 41: This is just a wrapper if another function so I don't think there's a lot of value to create another function here.

pkg/utils/aws.go
```diff
@@ -32,3 +37,40 @@ func DownloadS3ObjectToTempDir(ctx context.Context, downloader s3manageriface.Do
...
+		Region: aws.String(s3Region),
+		Credentials: credentials.NewStaticCredentials(
+			accessKeyId,
+			secretAccessKey,
+			""),
+	})
+	return s, err
+}
+
+func AddFileToS3Bucket(s *session.Session, s3Bucket string, dir string, fileName string) error {
```
wei-lee, line 52: to make it easier test this function, make the s3 api passed in. See https://github.com/integr8ly/delorean/blob/master/pkg/utils/aws.go#L15

wei-lee: You can use `s3iface.S3API` to pass the s3 interface in.

pkg/utils/zip.go
wei-lee: right, ok. Then the comment should be updated as well.

cmd/exportResults.go
```diff
@@ -0,0 +1,101 @@
...
+			awsSecretKey, err := requireValue(AWSSecretAccessKeyEnv)
+			if err != nil {
+				handleError(err)
+			}
+			bucket, err := requireValue(s3BucketEnv)
+			if err != nil {
+				handleError(err)
+			}
+
+			session, err := utils.CreateSession(awsKeyId, awsSecretKey, AWSDefaultRegion)
```
wei-lee, line 76: you can make the session passed in for testing.

cmd/exportResults.go
```diff
@@ -0,0 +1,101 @@
...
+			if err != nil {
+				handleError(err)
+			}
+			err = utils.AddFileToS3Bucket(session, bucket, f.testResultsDir, zipFile)
```
wei-lee, line 80: if you change this as I suggested, then you will be able to pass in the implementation for s3 to this function. Then you will be able to test.

cmd/exportResults.go
```diff
@@ -0,0 +1,101 @@
...
+			if err != nil {
+				handleError(err)
+			}
+			fmt.Println("Exported.")
```
wei-lee, line 84: Take a look at how this cmd is tested: https://github.com/integr8ly/delorean/blob/master/cmd/reportPortalImport.go. To test this func, you need to:

1. Add a new struct for the cmd. You can declare a few fields/dependencies that you will need to execute the function using interfaces. See https://github.com/integr8ly/delorean/blob/master/cmd/reportPortalImport.go#L40
2. In this command, initialise the dependencies the function will need: see https://github.com/integr8ly/delorean/blob/master/cmd/reportPortalImport.go#L114
3. Then you can provide mocked objects in the tests: https://github.com/integr8ly/delorean/blob/master/cmd/reportPortalImport_test.go#L127
4. You will be able to execute tests for most part of this function.

pkg/utils/zip.go
StanleyKaleta: updated

cmd/exportResults.go
StanleyKaleta: 👍

cmd/exportResults_test.go
```diff
@@ -0,0 +1,57 @@
...
+				},
+			},
+		},
+	}
+
+	for _, c := range cases {
+		t.Run(c.description, func(t *testing.T) {
+			tmpDir := "/tmp/export-unit-test/"
+			zipFile := "test-xyz.zip"
+			os.MkdirAll(tmpDir, os.ModePerm)
```
wei-lee, line 34: small thing but you can also use [ioutils.TempDir](https://golang.org/pkg/io/ioutil/#TempDir)

PR #101, opened 2020-06-24
DEL-308:update Openshift CI  release repo for major and minor release only

STEPHINRACHEL:
**What**
Update to the` openshift-ci-release` command to update the release repo only if the release is a major or minor one


cmd/openshiftCIRelease.go
```diff
@@ -343,25 +343,29 @@ func init() {
...
 			}
-			var intlyOperatorRepoDir string
-			if intlyOperatorRepoDir, err = c.DoIntlyOperatorUpdate(cmd.Context()); err != nil {
-				handleError(err)
-			}
-			if intlyOperatorRepoDir != "" {
-				fmt.Println("Remove temporary directory:", intlyOperatorRepoDir)
-				if err = os.RemoveAll(intlyOperatorRepoDir); err != nil {
+			if c.version.IsPreRelease() {
+				fmt.Println("Skipping the update to Openshift CI release repo as the release version is not a major or a minor one")
```
b1zzu, line 347: just one tip to keep the change smaller, if you put a `return` immideatlly after the `Println`you don't have to put the `else` stantment and add a new indedentation layer

PR #103, opened 2020-06-30
add get and set for relatedImages

StevenTobin:
If isGa and there are relatedImages found, replace any reference to the staging registries with the production registry.

*To Verify*
1. change images in relatedImages section in `pkg/utils/testdata/validManifests/3scale/0.4.0/3scale-operator.v0.4.0.clusterserviceversion.yaml` to point to staging registries
2. Run `./delorean ews process-csv-images -m pkg/utils/testdata/validManifests/3scale/ -g`
3. Verify all relatedImages now point to `registry.redhat.io`

### DIFF
--- a/pkg/utils/csv.go
+++ b/pkg/utils/csv.go
@@ -34,6 +34,12 @@ type csvName struct {
 	Name    string
 	Version semver.Version
 }
+
+type relatedImage struct {
+	Name  string
+	Image string
+}
+
 type csvNames []csvName
 
 func (c csvNames) Len() int           { return len(c) }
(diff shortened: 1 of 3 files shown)


pkg/utils/csv.go
```diff
@@ -374,6 +411,29 @@ func GetAndUpdateOperandImages(manifestDir string, extraImages []string, isGa bo
...
+				relatedImages[k].Image = strings.Replace(i.Image, "registry.stage.redhat.io", "registry.redhat.io", 1)
+			}
+
+			if strings.Contains(i.Image, "quay.io/integreatly/delorean") {
+				relatedImages[k].Image = strings.Replace(i.Image, "quay.io/integreatly/delorean", "registry.redhat.io", 1)
+			}
+
+			if strings.Contains(i.Image, "registry-proxy.engineering.redhat.com") {
+				relatedImages[k].Image = strings.Replace(i.Image, "registry-proxy.engineering.redhat.com", "registry.redhat.io", 1)
+			}
```
mikenairn, line 428: If it was ever to match quay.io/integreatly/delorean or registry-proxy.engineering.redhat.com it woudln't be enough to just replace the host.

This should never happen anyway, so i would just remove these for now

pkg/utils/csv.go
```diff
@@ -343,6 +379,7 @@ func GetAndUpdateOperandImages(manifestDir string, extraImages []string, isGa bo
 
 	var images = map[string]string{}
 	deployment, err := csv.GetOperatorDeploymentSpec()
+	relatedImages, err := csv.GetRelatedImages()
```
mikenairn, line 382: Not all CSVs will have related images, this looks like it will fail if it cant find the section in the CSV?

pkg/utils/csv.go
StevenTobin: Sure, I'll make this change

pkg/utils/csv.go
StevenTobin: It doesn't fail it just returns an empty array so I check if we found relatedImages later and act on that. So It doesn't hurt to do it like this but I could specifically check if you think it would be cleaner.

mikenairn: :+1:

pkg/utils/csv.go
```diff
@@ -374,6 +411,21 @@ func GetAndUpdateOperandImages(manifestDir string, extraImages []string, isGa bo
...
+				relatedImages[k].Image = strings.Replace(i.Image, "registry.stage.redhat.io", "registry.redhat.io", 1)
+			}
+		}
+
+		err = csv.SetRelatedImages(relatedImages)
+		if err != nil {
+			return nil, err
+		}
+	}
+
```
mikenairn, line 428: This task is to just update the registry URL, so this is fine for now. IN the future we will likely need to just process the relatedImages in the same way as all the rest since they will likely be used to actually install the product when we are using the new bundle format.

pkg/utils/csv.go
```diff
@@ -374,6 +431,21 @@ func GetAndUpdateOperandImages(manifestDir string, extraImages []string, isGa bo
 			}
 		}
 	}
+
+	// Check that relatedImages all point to the correct registry if isGa
+	if isGa && len(relatedImages) > 0 {
+		for k, i := range relatedImages {
+			if strings.Contains(i.Image, "registry.stage.redhat.io") {
+				relatedImages[k].Image = strings.Replace(i.Image, "registry.stage.redhat.io", "registry.redhat.io", 1)
```
mikenairn, line 439: This will make everything image right, even if it was value previously, we are changing the key?

StevenTobin: It will, the other csv's uniformly use image, and even in 3scale it switches half way through so image would seem to be the correct value. I can do it without changing if you'd prefer.

mikenairn: Lets just go with this for now.

PR #108, opened 2020-07-06
add cleanup cmd

wei-lee:
to verify:

* checkout the PR and build the cli
* Run:
   ```
    ./delorean report cleanup --config-file=./configurations/cleanup-reports.yaml
   ```
* You should see some outputs and some objects will be moved to the `archive/` sub dir in the bucket.
* Remove the objects back to the root dir manually.

### DIFF
--- a/cmd/cleanupReports.go
+++ b/cmd/cleanupReports.go
@@ -1,0 +1,207 @@
+package cmd
+
+import (
+	"context"
+	"fmt"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/aws/aws-sdk-go/service/s3/s3iface"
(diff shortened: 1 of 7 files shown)


b1zzu:
:eyes:

wei-lee:
@b1zzu are you able to review this PR yet?

cmd/cleanupReports.go
```diff
@@ -0,0 +1,207 @@
...
+		}
+		if ok {
+			fmt.Println(fmt.Sprintf("[%s] Object %s has matched tags and will be moved", bucket, *o.Key))
+			toCopy = append(toCopy, o)
+		} else {
+			fmt.Println(fmt.Sprintf("[%s] Skip object %s as it doesn't have the required tags", bucket, *o.Key))
+		}
+	}
+	copied := c.copyObjects(ctx, bucket, archiveFolderName, toCopy)
+	if err = c.deleteObjects(ctx, bucket, copied); err != nil {
```
b1zzu, line 136: maybe we should delete the object only if the copy succeeded otherwise we risk to permanently remove it

wei-lee: `copied` field will only contain objects that are copied successfully.

PR #110, opened 2020-07-08
make sure only list the objects in the root of a s3 bucket and ignore…

wei-lee:
… non-zip files


wei-lee:
@b1zzu this fixes a bug because there is now a subdirectory in the bucket.

PR #117, opened 2020-07-27
add datahub import

StevenTobin:
Add a `datahub-import` command to import data from a downtime-report to the datahub prometheus instance.

Test data can be seen here http://pushgateway-dh-prod-monitoring.cloud.datahub.psi.redhat.com/ and a grafana dashboard here https://grafana.datahub.redhat.com/dashboard/db/rhmi-product-downtime?orgId=1

To test:
1.Upload a `downtime-report.yaml` to an s3 bucket you have access to ( I can provide access if needed ) 
2.Run this command `./delorean report datahub-import -b <BUCKET-NAME> -p "http://pushgateway-redhat-rhmi-middleware-monitoring-operator.apps.stobin-prom.x0f5.s1.devshift.org/"`
3.See that the data has been pushed here http://pushgateway-redhat-rhmi-middleware-monitoring-operator.apps.stobin-prom.x0f5.s1.devshift.org

This PR includes a bump of prometheus/client_golang to 1.2.0 since that version fixes an upload error in pushgateway 1.10+ which was required.

The datahub grafana deployment does not use the grafana operator so the dashboard must be updated manually going forward and backed up in json form in the `templates/` directory


wei-lee:
👀

cmd/datahubImport.go
```diff
@@ -0,0 +1,217 @@
...
+
+func (c *datahubImportCmd) processReportFile(ctx context.Context, object *s3.Object) (interface{}, error) {
+	if !strings.HasPrefix(*object.Key, "downtime-report") {
+		fmt.Println(fmt.Sprintf("[%s] Skipping processing object", *object.Key))
+		return &struct{}{}, nil
+	}
+
+	fmt.Println(fmt.Sprintf("[%s] Start processing object", *object.Key))
+
+	_, err := c.s3.GetObjectTaggingWithContext(ctx, &s3.GetObjectTaggingInput{
```
wei-lee, line 143: it doesn't look like the tags are used. We should consider add a tag to the file once it's been processed so that we don't process the same file twice. Here then we should check if the object has the tag (which means it is already being processed).

We can then update [this configuration](https://github.com/integr8ly/delorean/blob/master/configurations/cleanup-reports.yaml) to cleanup the reports.

cmd/datahubImport.go
```diff
@@ -0,0 +1,217 @@
...
+
+	var name string
+	var count int
+	// Get version string
+	ver, err := utils.NewRHMIVersion(qr.Version)
+	if err != nil {
+		return nil, err
+	}
+	for _, i := range qr.Results {
+		if strings.Split(i.Name, "_")[0] != name {
```
wei-lee, line 175: I am not sure if this is right. By reading the code, I think the idea is that it will push a new metric to the gateway when the product changes. But statement will always be true the first time this is executed, so it will always push a metric  with a a value 0 to the gateway. Is that expected?

cmd/datahubImport.go
```diff
@@ -0,0 +1,217 @@
...
+			}
+		}
+
+		if len(i.v.String()) > 0 {
+			// The metric value comes with the query so split it to get the int value we care about
+			n, err := parseValue(i.v.String())
+			if err != nil {
+				return 0, err
+			}
+			count = count + n
```
wei-lee, line 201: so this is getting the total downtime for each product by adding then up? I don't think that's right. We shouldn't need to do that. We just need to send them as they are and if we need to add them up, we can do the aggregation via PromQL.

cmd/datahubImport.go
StevenTobin: Sure, I'll add that in, I saw that in the polarionImport but just forgot to add it here

cmd/datahubImport.go
StevenTobin: Removed the count and now push all metrics using product label to organise

cmd/datahubImport.go
StevenTobin: Fixed by pushing all metrics

cmd/datahubImport.go
```diff
@@ -0,0 +1,227 @@
...
+			if err != nil {
+				return 0, err
+			}
+			downtimeCount.Set(float64(count))
+
+		}
+		// push the metric
+		pusher := push.New(c.pushgateway, c.jobName)
+		pusher.Collector(downtimeCount).Grouping("product", name).Grouping("query", i.Name).Grouping("version", ver.String())
+		err = pusher.Push()
```
wei-lee, line 195: If I am not mistaken, rather than pushing multiple jobs (with the same name) for each result, we could just init the `pusher` object once at the beginning, and use the `Collector` method to add gauges to the `pusher`, and push the results at the end. This way you should end up with just 1 job in the gateway, but it will container multiple gauges.

PR #119, opened 2020-07-29
fix for pagerduty and deadmanssnitch secrets

jhellar:
Currently in staging pagerduty secret is created by hive, so the `oc` command to create the secret fails. This change makes sure that commands to create the monitoring secrets won't fail if already created by hive.

https://issues.redhat.com/browse/INTLY-9551


grdryn:
/lgtm

Some failing jobs there though

grdryn:
@mikenairn @laurafitzgerald could this get a review/approval please?

PR #121, opened 2020-07-29
DEL-357: Add verify-csv-version cmd

b1zzu:
**What**

This cmd will verify that the incoming operator version is not smaller than the version already on the integreatly-operator

**Try it**:

* you need to clone the integreatly-operator
* need to compile the Delorean cli

```
mkdir /tmp/3scale
./delorean ews extract-manifests --src-image "registry-proxy.engineering.redhat.com/rh-osbs/3scale-amp2-3scale-rhel7-operator-metadata:1.11.0-68" --extract-dir /tmp/3scale
./delorean ews verify-csv-version --incoming-manifests /tmp/3scale --current-manifests ../integreatly-operator/manifests/integreatly-3scale
```

the script should fail

### DIFF
--- a/cmd/verifyVersion.go
+++ b/cmd/verifyVersion.go
@@ -49,7 +49,7 @@ func doVerifyVersion(flags *verifyVersionFlags) error {
 		return err
 	}
 
-	fmt.Printf("comparing incoming CSV version %s with current version %s", incoming, current)
+	fmt.Printf("comparing incoming CSV version %s with current version %s\n", incoming, current)
 	o := incoming.Compare(current)
 	switch o {
 	case -1:
--- a/cmd/verifyVersion.go
+++ b/cmd/verifyVersion.go
@@ -1,0 +1,69 @@
+package cmd
+
+import (
+	"fmt"
+
+	"github.com/blang/semver"
+	"github.com/integr8ly/delorean/pkg/utils"
+	"github.com/spf13/cobra"
+)
+
+type verifyVersionFlags struct {
+	incomingManifests string
+	currentManifests  string
+}
+
+func init() {
+
+	flags := &verifyVersionFlags{}
+
+	cmd := &cobra.Command{
+		Use:   "verify-csv-version",
+		Short: "Compare the incoming CSV version from the incoming manifests dir with the current CSV version",
+		Run: func(cmd *cobra.Command, args []string) {
+			err := doVerifyVersion(flags)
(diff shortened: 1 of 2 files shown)


cmd/verifyVersion.go
```diff
@@ -0,0 +1,70 @@
...
+			err := doVerifyVersion(flags)
+			if err != nil {
+				handleError(err)
+			}
+		},
+	}
+
+	ewsCmd.AddCommand(cmd)
+
+	cmd.Flags().StringVar(&flags.incomingManifests, "incoming-manifests", "", "the manifests directory with the incoming CSV that must be bigger or equal then the current CSV")
```
mikenairn, line 33: We should be more consistent with the naming of the flags, pretty sure we use --manifest-dir elsewhere, and possibly other names, all for the same thing

cmd/verifyVersion.go
```diff
@@ -0,0 +1,70 @@
...
+	cmd.MarkFlagRequired("incoming-manifests")
+
+	cmd.Flags().StringVar(&flags.currentManifests, "current-manifests", "", "the manifests directory with the current CSV")
```
mikenairn, line 36: Ok, we have 2 the same, guess we need some difference in the naming then

b1zzu: yes,  I used two names to differentiate the manifests dir for the new package with the manifests dir in the integreatly-operator, I also know we used different names like `--directory` or `--manifest-dir`, so I use manifests in the name without dir to avid making a too long flag. Also an alternative was using `--left-manifests` and `--right-manifests` and just compare them, but then I would have to implement the failing logic in Jenkins

mikenairn:
Verified with 3scale:

```
$ ./delorean ews verify-csv-version --incoming-manifests /tmp/3scale --current-manifests /home/mnairn/go/src/github.com/integr8ly/integreatly-operator/manifests/integreatly-3scale
$ echo $?
0
$ ./delorean ews verify-csv-version --incoming-manifests /tmp/3scale --current-manifests /home/mnairn/go/src/github.com/integr8ly/integreatly-operator/manifests/integreatly-3scale
comparing incoming CSV version 0.5.1 with current version 0.5.4Error: the incoming operator version '0.5.1' is smaller than current version '0.5.4'
mnairn@deacon delorean (pr/121) $ echo $?
1
```

mikenairn:
It does what we need.

/lgtm

PR #122, opened 2020-07-30
Removal of pagerduty secret creation

trepel:
Fix for INTLY-9551

The Hive now creates the pagerduty secret, so it should not be created by the pipeline.

Also fixed bug in dms secret creation: `oc apply` does not have `--from-literal` flag, the `oc create` has to be used. This is basically a revert of https://github.com/integr8ly/delorean/commit/a658d684aba6b9cdb2a50042180df798a0521ecc

### DIFF
--- a/scripts/ocm/ocm.sh
+++ b/scripts/ocm/ocm.sh
@@ -148,9 +148,10 @@ install_rhmi() {
             --from-literal=port=587 \
             --from-literal=tls=true
     fi
-    oc --kubeconfig "${CLUSTER_KUBECONFIG_FILE}" apply secret generic redhat-rhmi-pagerduty -n ${RHMI_OPERATOR_NAMESPACE} \
+    oc --kubeconfig "${CLUSTER_KUBECONFIG_FILE}" get secret redhat-rhmi-pagerduty -n ${RHMI_OPERATOR_NAMESPACE} \
+        || oc --kubeconfig "${CLUSTER_KUBECONFIG_FILE}" create secret generic redhat-rhmi-pagerduty -n ${RHMI_OPERATOR_NAMESPACE} \
         --from-literal=serviceKey=dummykey
-    oc --kubeconfig "${CLUSTER_KUBECONFIG_FILE}" apply secret generic redhat-rhmi-deadmanssnitch -n ${RHMI_OPERATOR_NAMESPACE} \
+    oc --kubeconfig "${CLUSTER_KUBECONFIG_FILE}" create secret generic redhat-rhmi-deadmanssnitch -n ${RHMI_OPERATOR_NAMESPACE} \
         --from-literal=url=https://dms.example.com
 
     if [[ "${PATCH_CR_AWS_CM}" == true ]]; then


scripts/ocm/ocm.sh
```diff
@@ -148,9 +148,7 @@ install_rhmi() {
             --from-literal=port=587 \
             --from-literal=tls=true
     fi
-    oc --kubeconfig "${CLUSTER_KUBECONFIG_FILE}" apply secret generic redhat-rhmi-pagerduty -n ${RHMI_OPERATOR_NAMESPACE} \
-        --from-literal=serviceKey=dummykey
-    oc --kubeconfig "${CLUSTER_KUBECONFIG_FILE}" apply secret generic redhat-rhmi-deadmanssnitch -n ${RHMI_OPERATOR_NAMESPACE} \
+    oc --kubeconfig "${CLUSTER_KUBECONFIG_FILE}" create secret generic redhat-rhmi-deadmanssnitch -n ${RHMI_OPERATOR_NAMESPACE} \
```
mikenairn, line 151: Did we not just switch this from using create to apply?

trepel: apply does not have `--from-literal` flag so it can't be used

mikenairn: ok, so the previous change was never tested.

mikenairn:
@psturc Do you want to check this, you'd be more familiar with the current usage of these scripts that me.

trepel:
@psturc I tried with addon-flow. It failed due to https://issues.redhat.com/browse/INTLY-9617 but it got past this step:

https://master-jenkins-csb-intly.cloud.paas.psi.redhat.com/job/Integreatly/job/rhmi-install-addon-flow/214/parameters/

scripts/ocm/ocm.sh
trepel: yes, that was my fault, I misinterpreted the pipeline log. I thought that the pipeline passed that step and only failed later but it was not true. Since the pipeline was unstable for various other reasons it took few hours to me to realize that this needs to be reverted.

trepel:
@psturc or whoever, what about versioning? If Hive creates the secret for 2.6.0, this has to be merged. But if Hive does not create secret for 2.5.0, we need to version this repo to be able to provision previous releases, am I right?

psturc:
sorry, somehow I managed to miss this PR yesterday.

  > ...we need to version this repo to be able to provision previous releases, am I right?

yes, either we can create a special `2.5.0` and `2.6.0` tags in this repo, or you can actually just check whether the secret doesn't exist before creating it, which would be safe for all versions of RHMI so far

e.g.

```
oc --kubeconfig "${CLUSTER_KUBECONFIG_FILE}" get secret redhat-rhmi-pagerduty -n ${RHMI_OPERATOR_NAMESPACE} \
|| oc --kubeconfig "${CLUSTER_KUBECONFIG_FILE}" create secret generic redhat-rhmi-pagerduty -n ${RHMI_OPERATOR_NAMESPACE} \
        --from-literal=serviceKey=dummykey
```

trepel:
@psturc Your suggestions has been accepted, would you mind having another look?
@pawelpaszki fyi - this should fix the nightly pipeline

Running addon-flow from this PR to fully verify this:
https://master-jenkins-csb-intly.cloud.paas.psi.redhat.com/job/Integreatly/job/rhmi-install-addon-flow/221/

psturc:
/lgtm
I will approve it if the pipeline succeeds

PR #123, opened 2020-07-31
add set-product-version command

StevenTobin:
Adds set-product-version command to set the productVersion and operatorVersion in the rhmi_types file to either "CHANGEME" or the discovered operator version.

https://issues.redhat.com/browse/DEL-213

To Verify

1. Checkout this branch and build
2. Run `./delorean ews set-product-version -d cmd/testdata/manageRHMITypes/rhmi_types -p 3scale -v 9.9.9`
3. Verify that the `OperatorVersion3Scale` var has been changed to `9.9.9` and that the `Version3Scale` var has not been changed
4. Run `./delorean ews set-product-version -d cmd/testdata/manageRHMITypes/rhmi_types -p 3scale -v 9.10.0`
5. Verify that the `OperatorVersion3Scale` var has been changed to `9.10.0` and that the `Version3Scale` var has been changed to `CHANGEME`

### DIFF
--- a/cmd/manageRHMITypes.go
+++ b/cmd/manageRHMITypes.go
@@ -1,0 +1,126 @@
+package cmd
+
+import (
+	"fmt"
+	"github.com/spf13/cobra"
+	"golang.org/x/mod/semver"
+	"io/ioutil"
+	"os"
+	"regexp"
+	"strings"
+)
+
+type manageTypesCmdOptions struct {
+	directory string
+	product   string
+	version   string
+}
+
+func init() {
+
+	flags := &manageTypesCmdOptions{}
+
+	cmd := &cobra.Command{
+		Use:   "set-product-version",
+		Short: "Sets the operator and product version for a product in the rhmi_types file",
+		Run: func(cmd *cobra.Command, args []string) {
+			err := SetVersion(flags.directory, flags.product, flags.version)
+			if err != nil {
+				handleError(err)
+			}
+		},
+	}
+
+	ewsCmd.AddCommand(cmd)
+
+	cmd.Flags().StringVarP(&flags.directory, "directory", "d", "", "Path to rhmi_types file")
+	cmd.MarkFlagRequired("directory")
+
+	cmd.Flags().StringVarP(&flags.product, "product", "p", "", "The product name")
+	cmd.MarkFlagRequired("product")
+
+	cmd.Flags().StringVarP(&flags.version, "version", "v", "", "The desired version")
+	cmd.MarkFlagRequired("version")
+}
+
+func SetVersion(directory string, product string, version string) error {
+	product = PrepareProductName(product)
+	fmt.Println(fmt.Sprintf("setting version of product %s to %s", product, version))
+	read, err := os.Open(directory)
+	if err != nil {
+		return err
+	}
+	bytes, err := ioutil.ReadAll(read)
+	if err != nil {
+		return err
+	}
+	out := ParseVersion(bytes, product, version)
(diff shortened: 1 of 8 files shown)


cmd/manageRHMITypes.go
```diff
@@ -0,0 +1,126 @@
...
+	case "codeready-workspaces":
+		product = "CodeReadyWorkspaces"
+	case "fuse-online":
+		product = "FuseOnline"
+	case "rhsso":
+		product = "RHSSO"
+	}
+
+	return product
+}
```
mikenairn, line 85: It's a pity we have to hard code product names like this, but i guess there is no easy way to convert the name from what the pipeline knows to the Golang variable name.

Longer term it would probably be better to use a different machine readable file to store the values in the operator, rather than manipulating the go code in this way, but considering the changes comeing down the line, probably not worth thinking about at the moment.

cmd/manageRHMITypes.go
```diff
@@ -0,0 +1,126 @@
...
+			err := SetVersion(flags.directory, flags.product, flags.version)
+			if err != nil {
+				handleError(err)
+			}
+		},
+	}
+
+	ewsCmd.AddCommand(cmd)
+
+	cmd.Flags().StringVarP(&flags.directory, "directory", "d", "", "Path to rhmi_types file")
```
mikenairn, line 36: Not sure "directory" makes sense for the name here see https://gitlab.cee.redhat.com/integreatly-qe/ci-cd/-/merge_requests/256#note_1539372

cmd/manageRHMITypes.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	// if it's not a minor release only change the OperatorVersion<product> var
+	// if the version is 0.x.x then semver returns empty string so account for it
+	if c != 0 || currentOperatorMajor == "" && newOperatorMajor == "" {
+		out = strings.Replace(string(in), ovs, version, 1)
+	}
+
+	return out
+}
```
mikenairn, line 126: Looks like this all assumes that the product version and operator version is the same, is that correct?

Product versions and Operator version need to be handled differently as they are not necessarily the same. see https://github.com/integr8ly/integreatly-operator/commit/1ab6531ec42ebb6a29a96dbb4ffc33b2f98e6390

mikenairn: OK, looking at it again, we never try and set the product version, but base everything off the operator version and only update the product version to CHANGE. Guess it might be fine.

StevenTobin: I'll change that, I use product a few times where I actually mean operator, it is confusing looking at it now

cmd/manageRHMITypes.go
StevenTobin: Changed to filepath to make it clearer

cmd/manageRHMITypes_test.go
```diff
@@ -0,0 +1,156 @@
...
+		}
+	}
+
+	if majmin == "minor" {
+		if ovs != version {
+			t.Errorf("error found incorrect version string, wanted = %s, got %s", ovs, version)
+		}
+		if pvs != "CHANGEME" {
+			t.Errorf("error found incorrect version string, wanted = CHANGEME got %s", pvs)
+		}
```
mikenairn, line 152: Nothing should be getting set to CHANGEME now, is this test doing anything?

StevenTobin: No, it's not, I'll remove

PR #125, opened 2020-08-05
[WIP] Extend tag-repository cmd to support multiple repositories

b1zzu:
Note: This is only a POC, haven't actually tried if it works

**Usage:**
```
./delorean release tag-release --config-file ./configurations/tag-release.yaml
```

### DIFF
--- a/configurations/tag-release.yaml
+++ b/configurations/tag-release.yaml
@@ -1,0 +1,19 @@
+---
+version: 0.0.0
+branch: master
+wait: true
+waitInterval: 5
+waitMax: 90
+repositories:
+  - name: integreatly-operator
+    githubRepo:
+      owner: integr8ly
+      repo: integreatly-operator
+    quayRepos:
+      - integreatly/integreatly-operator
+      - integreatly/integreatly-operator-test-harness
+  - name: integreatly-help
+    skipPreRelease: true
+    githubRepo:
+      owner: RHCloudServices
+      repo: integreatly-help
--- a/cmd/tagRelease.go
+++ b/cmd/tagRelease.go
@@ -3,124 +3,169 @@ package cmd
 import (
 	"context"
 	"fmt"
+	"net/http"
+	"sort"
+	"strings"
+	"time"
+
 	"github.com/google/go-github/v30/github"
 	"github.com/integr8ly/delorean/pkg/quay"
 	"github.com/integr8ly/delorean/pkg/services"
 	"github.com/integr8ly/delorean/pkg/utils"
 	"github.com/spf13/cobra"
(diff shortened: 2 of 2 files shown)


configurations/tag-release.yaml
```diff
@@ -0,0 +1,19 @@
+---
+version: 0.0.0
+branch: master
+wait: true
+waitInterval: 5
+waitMax: 90
```
wei-lee, line 6: the above options are dynamic and they are likely to change, so it is better to pass them in as command line options, rather than using the config file.

configurations/tag-release.yaml
```diff
@@ -0,0 +1,19 @@
...
+repositories:
```
wei-lee, line 7: the following options are unlikely to change so it's better to use a configuration file to specify them.

configurations/tag-release.yaml
```diff
@@ -0,0 +1,19 @@
...
+      owner: integr8ly
+      repo: integreatly-operator
+    quayRepos:
+      - integreatly/integreatly-operator
+      - integreatly/integreatly-operator-test-harness
+  - name: integreatly-help
+    skipPreRelease: true
+    githubRepo:
+      owner: RHCloudServices
+      repo: integreatly-help
```
wei-lee, line 19: Based on my comments above, I think it's better to just use the configuration file to just specify the repos. This way the configuration file can be used by multiple commands, if they all need access to repo information.
Perhaps something like this:

```
---
repositories:
  - name: integreatly-operator
    githubRepo:
      owner: integr8ly
      repo: integreatly-operator
    quayRepos:
      - integreatly/integreatly-operator
      - integreatly/integreatly-operator-test-harness
    tagImages: true
    # can add additional options, or even command-specific ones (not sure if that's a good idea tbh)
    options:
      tagRelease: # options for the `tag-release` command
  - name: integreatly-help
    skipPreRelease: true
    githubRepo:
      owner: RHCloudServices
      repo: integreatly-help
    tagImages: false
    options:
      ...
```

configurations/tag-release.yaml
b1zzu: Yes, I was thinking to do it for the `version` and `branch` but I think it makes sense to keep the wait options in the config file, wdyt?

b1zzu: also regarding the `branch` wouldn't make sense to generate it from the version:
```
if version.IsPatch {
  branch = version.MajorMinor
}
```

wei-lee: if we want to keep the wait options in the config file, that's fine. But I think it probably should be specific to a repo for a specific command. As I suggested in the comment below.

PR #126, opened 2020-08-10
Del 386 add to group

wei-lee:
https://issues.redhat.com/browse/DEL-386

See commit f56101543046c22515d492e6d27d70283a298f9a


STEPHINRACHEL:
@wei-lee Tried running the product-tests command but got error as below; 
`[Prepare] Create namespace test-functional
[Prepare] Create serviceAccount cluster-admin-sa
[Prepare] Create ClusterRoleBinding for the service account
[Prepare] Add ServiceAccount to group osd-sre-cluster-admins
Error: Group.user.openshift.io "osd-sre-cluster-admins" is invalid: user[0]: Invalid value: "system:serviceaccount:test-functional:cluster-admin-sa": may not contain ":"`

wei-lee:
this actually is not needed. Closing

PR #127, opened 2020-08-11
osd cluster cidr range change (v4.4+)

psturc:
https://issues.redhat.com/browse/INTLY-9514

Verified on 4.4 BYOC cluster (public & private)

### DIFF
--- a/templates/ocm/cluster-template.json
+++ b/templates/ocm/cluster-template.json
@@ -4,10 +4,10 @@
     "multi_az": false,
     "name": "my_cluster_name",
     "network": {
-      "machine_cidr": "10.11.128.0/23",
+      "machine_cidr": "10.11.128.0/25",
       "service_cidr": "10.11.0.0/18",
       "pod_cidr": "10.11.64.0/18",
-      "host_prefix": 23
+      "host_prefix": 26
     },
     "nodes" : {
       "compute" : 4,


pawelpaszki:
/lgtm
will let others take a look and can approve then

trepel:
@psturc Hm, I was release test coordinator for 2.6.0 (july 27 - august 5), we tested on OSD 4.4.11 and I am not aware of using this PR for the https://master-jenkins-csb-intly.cloud.paas.psi.redhat.com/job/Integreatly/job/rhmi-install-addon-flow/ pipeline.

Based on what's written in https://github.com/RHCloudServices/integreatly-help/blob/master/qe-guides/2.x-release-testing-workflow.md#public-osd4-byoc-cluster--rhmi-installation-addon-flow it sounds like it shouldn't have worked, but it had.

So I am slightly confused.

psturc:
@trepel ok, when looking again at the note in testing-workflow doc you've linked, I realize it might be a little confusing.

I wrote there that "we needed to update the cluster networking template used for provisioning of OSD testing clusters", but the update is actually not **required** in order to successfully provision OSD 4.4.x cluster. With old network config, you would still be able to provision any version of OSD cluster. The new change I'm introducing in this PR is just an update to the **minimal network configuration** of a cluster that customer can ask for.

You can find more details in [this JIRA](https://issues.redhat.com/browse/INTLY-9514) and linked SDA ticket.

Hope this clears out the confusion. Let me know if there's anything else unclear.

trepel:
@psturc ah, ok, it makes sense now. Feel free to approve (are you the only approver?).

Once this is merged we should revert https://github.com/RHCloudServices/integreatly-help/pull/653/files, correct?
Also, shouldn't we update the test case: https://github.com/integr8ly/integreatly-operator/blob/master/test-cases/tests/upgrade/n03-verify-that-the-minimal-cidr-block-is-configured-correctly.md accordingly?

psturc:
@trepel  I would rather wait with merging this PR, because we still plan to test RHMI upgrade from version 2.4.0, which should be installed on OpenShift 4.3.x.
Once we are done with it, I think we can merge this PR and update the guide and the test case you've mentioned.

cc @b1zzu: should there be a note about this in the current test plan?

b1zzu:
@psturc I'm aiming to install the 2.4.0 on OSD 4.4.6 but I'm not yet sure on how to do it, if this PR can complicate things then yes is better to wait

Boomatang:
@psturc Doing some house keeping. This PR is very only and we no longer use osd 4.4. Do you think this PR can be closed?

PR #131, opened 2020-08-12
add creation of secret and switch to upstream test image

StevenTobin:
We are switching to use the test image maintained by the 3scale team. To facilitate that we need to allow access to the quay repository which contains the test image by creating a dockercfg secret with the auth details and adding it to the job template with an ImagePullSecrets field.

### DIFF
--- a/cmd/productTests.go
+++ b/cmd/productTests.go
@@ -140,7 +140,7 @@ func (c *runTestsCmd) run(ctx context.Context) error {
 	if _, err = utils.CreateClusterRoleBinding(c.clientset, sa, "cluster-admin", *owner); err != nil {
 		return err
 	}
-	err = utils.CreateSecret(c.clientset, secretName, "quay.io", c.namespace, username, password)
+	err = utils.CreateDockerSecret(c.clientset, secretName, "quay.io", c.namespace, username, password)
 	if err != nil {
 		return err
 	}
--- a/cmd/productTests.go
+++ b/cmd/productTests.go
@@ -27,6 +27,7 @@ const (
 	testJobBackoffLimit = 0
 	defaultNamespace    = "rhmi-product-tests"
 	serviceAccountName  = "cluster-admin-sa"
+	secretName          = "rh-integration-auth"
 )
 
 type TestContainer struct {
@@ -117,6 +118,14 @@ func (c *runTestsCmd) run(ctx context.Context) error {
 	var ns *v1.Namespace
 	var sa *v1.ServiceAccount
 	var err error
+	username, err := requireValue(RHIntegrationUsername)
+	if err != nil {
+		return err
+	}
+	password, err := requireValue(RHIntegrationPassword)
+	if err != nil {
+		return err
+	}
 	fmt.Println("[Prepare] Create namespace", c.namespace)
 	if ns, err = utils.CreateNamespace(c.clientset, c.namespace); err != nil {
 		return err
@@ -129,6 +138,10 @@ func (c *runTestsCmd) run(ctx context.Context) error {
 	gvk := schema.FromAPIVersionAndKind("v1", "namespace")
 	owner := metav1.NewControllerRef(ns, gvk)
 	if _, err = utils.CreateClusterRoleBinding(c.clientset, sa, "cluster-admin", *owner); err != nil {
+		return err
+	}
+	err = utils.CreateSecret(c.clientset, secretName, "quay.io", c.namespace, username, password)
+	if err != nil {
 		return err
 	}
 	var wg sync.WaitGroup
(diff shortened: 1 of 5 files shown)


pkg/utils/k8s.go
```diff
@@ -108,6 +109,32 @@ func CreateClusterRoleBinding(client kubernetes.Interface, sa *v1.ServiceAccount
 	return r, nil
 }
 
+func CreateSecret(client kubernetes.Interface, secretName string, registry string, namespace string, username string, password string) error {
```
wei-lee, line 112: small thing but the name of the function can be a bit more specific - it is creating a docker secret, rather than just a generic secret.

StevenTobin: Yeah good point, pushed a change

cmd/productTests.go
```diff
@@ -117,6 +118,14 @@ func (c *runTestsCmd) run(ctx context.Context) error {
...
 	var sa *v1.ServiceAccount
 	var err error
+	username, err := requireValue(RHIntegrationUsername)
+	if err != nil {
+		return err
+	}
+	password, err := requireValue(RHIntegrationPassword)
+	if err != nil {
+		return err
+	}
```
wei-lee, line 128: while this works, the downside is that if we need to support another test container image from a different private org, we will have to add another envvar again. Also I don't think these should be required if we want to run test containers that are not from this org (e.g. we are using the same cli to run the integreatly operator test harness image, but it will not work now once this change is merged).

I suggest we move this option to the configuration file. Each test container image can provide an additional value for an env var that will store the docker username and password, or even registry. 

For example, the configuration will be something like this:

```
---
tests:
  - name: 3scale-test
    image: quay.io/integreatly/3scale-py-testsuite:0.1.0
    timeout: 3600
    imagePullSecretEnvVar: "RH_INTEGRATION_CREDENTIAL"
    envVars:
      - name: NAMESPACE
        value: redhat-rhmi-3scale
```

In the cli, for each container, if this value is set, the cli will then look for the value of this env var, and create a secret from it and use it for the test job. If the image pull didn't exist, log the error and ignore the container. The value for the env var should be a stringified json object:

```
export RH_INTEGRATION_CREDENTIAL='{"username":"xxxx", "password":"xxxx", "registry":"quay.io"}'
```

Then in the cli, if the env var is found, we should be able to parse it and get the value we need and create the pull secret.

Then in the jenkins pipeline, we just need to ensure the env var required by the 3scale test container is set.

If the future, if another test container requires a different credential, all we need to do is to ensure the required env var is set.

wdyt?

StevenTobin: Yep that makes complete sense. I've pushed a change to reflect that flow.

cmd/productTests.go
```diff
@@ -131,8 +133,22 @@ func (c *runTestsCmd) run(ctx context.Context) error {
 	if _, err = utils.CreateClusterRoleBinding(c.clientset, sa, "cluster-admin", *owner); err != nil {
 		return err
 	}
+	if err != nil {
+		return err
+	}
```
wei-lee, line 138: is this change actually required?

cmd/productTests.go
```diff
@@ -265,6 +280,10 @@ func getTestContainerJob(namespace string, t *TestContainer) *batchv1.Job {
 					},
 					RestartPolicy:      "Never",
 					ServiceAccountName: serviceAccountName,
+					ImagePullSecrets: []v1.LocalObjectReference{{
+						Name: strings.ToLower(strings.ReplaceAll(t.ImagePullSecret, "_", "-")),
```
wei-lee, line 284: this function appeared twice, better to move it to a function.

pkg/utils/k8s.go
```diff
@@ -108,6 +109,31 @@ func CreateClusterRoleBinding(client kubernetes.Interface, sa *v1.ServiceAccount
...
 
+func CreateDockerSecret(client kubernetes.Interface, secretName string, namespace string, authstring string) error {
+	type RegistryAuth struct {
+		Username string `json:"username"`
+		Password string `json:"password"`
+		Registry string `json:"registry"`
+	}
+	type DockerAuth map[string]RegistryAuth
+	r := RegistryAuth{}
+	err := json.Unmarshal([]byte(authstring), &r)
```
wei-lee, line 120: small thing, needs to handle err here

pkg/utils/k8s.go
```diff
@@ -108,6 +109,31 @@ func CreateClusterRoleBinding(client kubernetes.Interface, sa *v1.ServiceAccount
...
+	auth, err := json.Marshal(DockerAuth{r.Registry: r})
```
wei-lee, line 121: and err here

cmd/productTests.go
StevenTobin: It isn't, I removed the secret creation from here and forgot to remove this, removing

cmd/productTests.go
StevenTobin: Sure, moved to a function

PR #135, opened 2020-08-27
DEL-368: Upload downtime reports to s3

StanleyKaleta:
example cmds:
```
oc login ...
export DELOREAN_AWS_ACCESS_KEY_ID
export DELOREAN_AWS_SECRET_ACCESS_KEY
./delorean report query-report --config-file `pwd`/configurations/downtime-report-config.yaml -o `pwd`/downtime-report --s3-bucket=rhmi-test-data-storage-tf56 --pipeline-name="upgrade-54"
```

### DIFF
--- a/cmd/queryReport.go
+++ b/cmd/queryReport.go
@@ -4,6 +4,11 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/s3/s3manager"
+	"github.com/aws/aws-sdk-go/service/s3/s3manager/s3manageriface"
 	"github.com/integr8ly/delorean/pkg/services"
 	"github.com/integr8ly/delorean/pkg/utils"
 	routeclientv1 "github.com/openshift/client-go/route/clientset/versioned/typed/route/v1"
@@ -12,6 +17,7 @@ import (
 	"github.com/prometheus/common/config"
 	"github.com/prometheus/common/model"
 	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/tools/clientcmd"
 	"math"
@@ -131,6 +137,8 @@ type queryReportCmd struct {
 	config     *queryReportConfig
 	queryRange queryRange
 	version    string
+	bucket     string
+	uploader   s3manageriface.UploaderAPI
 }
 
 type queryReportCmdFlags struct {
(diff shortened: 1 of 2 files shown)


cmd/queryReport.go
```diff
@@ -163,7 +190,7 @@ func init() {
 			}
 		},
 	}
-	pipelineCmd.AddCommand(cmd)
+	reportCmd.AddCommand(cmd)
```
wei-lee, line 193: from the description of the commands, I think it's better to keep this command as a subcommand for `pipeline`. The reporting commands are mainly for importing data into Polarion or reportportal. Here we are still mainly about generating the report and upload to s3.

cmd/queryReport.go
```diff
@@ -154,7 +167,21 @@ func init() {
 			if err != nil {
 				handleError(err)
 			}
-			c, err := newQueryReportCmd(kubeConfig, f)
+			awsKeyId, err := requireValue(AWSAccessKeyIDEnv)
+			if err != nil {
+				handleError(err)
+			}
+			awsSecretKey, err := requireValue(AWSSecretAccessKeyEnv)
```
wei-lee, line 174: these keys should only be required if the `s3bucket` flag is set. It should be possible to run the command without uploading the report.

cmd/queryReport.go
```diff
@@ -154,7 +167,21 @@ func init() {
...
+			if err != nil {
+				handleError(err)
+			}
+
+			s := session.Must(session.NewSession(&aws.Config{
+				Region:      aws.String(AWSDefaultRegion),
+				Credentials: credentials.NewStaticCredentials(awsKeyId, awsSecretKey, ""),
+			}))
```
wei-lee, line 182: this should only be needed when it is required to upload the report.

cmd/queryReport.go
```diff
@@ -246,6 +292,22 @@ func (c *queryReportCmd) run(ctx context.Context) error {
 		return err
 	}
 	fmt.Println("Report is generated:", outputFile)
+
+	zippedDir := c.outputDir + "/"
+	// zip the results
+	fmt.Println("Generating... " + zippedDir + c.zipFile)
+	err = utils.ZipFolder(zippedDir, zippedDir+c.zipFile)
```
wei-lee, line 299: there is no need to zip the folder here. There is only one file that needs to be uploaded. If you want to zip the files here, make sure you will need to unzip them [here](https://github.com/integr8ly/delorean/blob/master/cmd/datahubImport.go#L161)

wei-lee (changes requested):
see my comments above

cmd/queryReport.go
StanleyKaleta: ok

cmd/queryReport.go
StanleyKaleta: @wei-lee updated PR

wei-lee (approved):

PR #137, opened 2020-09-09
Revert "remove datahub tag from cleanup"

StanleyKaleta:
Reverts integr8ly/delorean#133

### DIFF
--- a/configurations/cleanup-reports.yaml
+++ b/configurations/cleanup-reports.yaml
@@ -6,3 +6,5 @@ configs:
         value: true
       - key: rp
         value: true
+      - key: datahub
+        value: true


configurations/cleanup-reports.yaml
```diff
@@ -5,4 +5,6 @@ configs:
       - key: polarion
         value: true
       - key: rp
+        value: true
+      - key: datahub
         value: true
```
wei-lee, line 10: this won't work. The relationship between all the tags for a bucket is "and", which means if you do it this way, a file will only be archived if it has all 3 tags. We don't want that.

You need to create another item under the `configs` list, and specified the tags. You can use the same bucket.

wei-lee (changes requested):
see my comment

configurations/cleanup-reports.yaml
StanleyKaleta: ack

StanleyKaleta:
@wei-lee created new PR for this: https://github.com/integr8ly/delorean/pull/154

PR #138, opened 2020-09-14
fix: remove smtp secret creation/deletion

psturc:
https://issues.redhat.com/browse/INTLY-9992

Sendgrid service should be now fully managing creating/deleting smtp creds + secret, so we no longer need to do this in a pipeline.

### DIFF
--- a/scripts/ocm/ocm.sh
+++ b/scripts/ocm/ocm.sh
@@ -157,13 +157,6 @@ delete_cluster() {
     cluster_id=$(get_cluster_id)
     infra_id=$(get_infra_id)
 
-    # Delete SMTP API key if SENDGRID_API_KEY is defined and infra_id is not empty
-    # infra_id would be empty if the cluster was not fully provisioned
-    if [[ -n "${SENDGRID_API_KEY:-}" ]] && [[ -n "${infra_id:-}" ]]; then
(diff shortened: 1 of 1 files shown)


scripts/ocm/ocm.sh
```diff
@@ -281,24 +274,6 @@ create_secrets() {
     local secrets
     secrets=$(oc --kubeconfig "${CLUSTER_KUBECONFIG_FILE}" get secrets -n "${RHMI_OPERATOR_NAMESPACE}" || true)
 
-    # If SMTP Secret is not present in RHMI operator namespace
-    if ! grep -q smtp <<< "${secrets}"; then
-        if [[ -n "${SENDGRID_API_KEY:-}" ]]; then
-            infra_id=$(get_infra_id)
-            echo "Creating SMTP secret with Sendgrid API key with ID: ${infra_id}"
-            smtp-service create "${infra_id}" | oc --kubeconfig "${CLUSTER_KUBECONFIG_FILE}" create -n "${RHMI_OPERATOR_NAMESPACE}" -f - \
```
psturc, line 289: If the secret doesn't get created automatically, it is most probably an issue with Sendgrid service we should address and a removal of this code should discover the issue.

psturc:
ping @pawelpaszki

PR #140, opened 2020-09-16
ocm: add support for multi_az cluster config

psturc:
### JIRA
https://issues.redhat.com/browse/MGDAPI-52

### Comments
* multi AZ configuration is based on the config in this [doc](https://docs.google.com/document/d/1nWe3Kvwg-53tKEz72mqcNs8u1AenYzjEAfkDb3W16kM/edit#heading=h.d3b3khyts47) created from [this spike](https://issues.redhat.com/browse/MGDAPI-12)
* I will link the jenkins job build that uses this configuration for creating a cluster

### Verification
1. Clone this branch

#### Config for Single AZ cluster
2. Run `make ocm/cluster.json | jq -r '.multi_az, .nodes'`
3. The output should look as following (the default single AZ configuration):
```
false
{
  "compute": 4,
  "compute_machine_type": {
    "id": "m5.xlarge"
  }
}
```
4. You should get the same output after running `make ocm/cluster.json MULTI_AZ=false | jq -r '.multi_az, .nodes'`

#### Config for Multi AZ cluster
2. Run `make ocm/cluster.json MULTI_AZ=true | jq -r '.multi_az, .nodes'`
3. The output should look as following (multi AZ configuration):
```
true
{
  "compute": 9,
  "compute_machine_type": {
    "id": "r5.xlarge"
  }
}
```

PR #141, opened 2020-09-16
doc: add note in cluster creation doc on multiple azs

darahayes:
Ping @psturc just some docs change to go along with your recent changes in #140

### DIFF
--- a/docs/ocm/README.md
+++ b/docs/ocm/README.md
@@ -36,6 +36,14 @@ export AWS_SECRET_ACCESS_KEY=<REPLACE_ME>
 export BYOC=true
 ```
 
+**Multiple AZ**
+
+If you want your BYOC cluster to span multiple availability zones then set the `MULTI_AZ` environment variable:
+
+```
+export MULTI_AZ=true
+```
+
 4. Create cluster template: `make ocm/cluster.json`
 
 This command will generate `ocm/cluster.json` file with generated cluster name. This file will be used as a template to create your cluster via OCM CLI.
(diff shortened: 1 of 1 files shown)


darahayes:
My plan is to then create a doc in the `integreatly-operator` repo that points to this doc for cluster creation and then documents how to install managed-api on top of it.

psturc:
thanks @darahayes, that's great, I'm always forgetting about documentation :)

> My plan is to then create a doc in the `integreatly-operator` repo that points to this doc for cluster creation and then documents how to install managed-api on top of it.

would it also make sense to modify [this doc](https://github.com/RHCloudServices/integreatly-help/blob/master/guides/ocp4/osd_dev_cluster.md) in the integreatly-help repo?

/lgtm
/approve

darahayes:
@psturc to me that doc you linked in the integreatly-help repo almost looks like a duplicate of the one here. Would it make sense to maybe remove the steps in there (creating cluster definition files) and just linking to this doc? In any case, both docs are only about getting a cluster and do not mention actually installing integreatly/managed-api/whatever. I think for actual managed-api/rhmi installation, we should probably document that in the integreatly-operator repo.

psturc:
@darahayes it makes sense to me to add a link for `docs/ocm/README.md` to the doc in integreatly-help repo, instead of documenting the same thing there.

The guide in this repo actually mentions the installation of RHMI via **addon**, but adding a guide for managed-api/rhmi installation (from master branch) to integreatly-operator repo seems reasonable to me. If you're going to create that guide, I think it would be good to also add a link to it to `docs/ocm/README.md` as well.

PR #142, opened 2020-09-18
MGDAPI-5: Add make target to install Managed API addon

sergioifg94:
# Description

> Link to JIRA: https://issues.redhat.com/browse/MGDAPI-5

_Changes required for the Managed API addon pipeline: https://gitlab.cee.redhat.com/integreatly-qe/ci-cd/-/merge_requests/307_

Add new `make ocm/install/managed-api-addon` target, it runs a new function in the
`ocm.sh` script that, similar to `ocm/install/rhmi-addon`, installs the addon in the
cluster and waits for the installation to finish, but uses the Managed API addon, and
relies on the products phase to decide when the installation was completed, as RHMI
relies on the Solution Explorer.

## Verification steps

1. Provision an OSD cluster and log into it
2. Save the cluster details into a file:
    ```sh
    ocm get /api/clusters_mgmt/v1/clusters/<CLUSTER ID> | jq -r  > ocm/cluster-details.json
    ```
3. Copy your kube config into the `ocm` folder
    ```sh
    # Replace "~/.kube/config" if you use a different location
    cp ~/.kube/config ocm/cluster.kubeconfig
    ```
4. Run the make target
    ```sh
    make ocm/install/managed-api-addon
    ```
5. Verify that the command works and the installation completes

### DIFF
--- a/scripts/ocm/ocm.sh
+++ b/scripts/ocm/ocm.sh
@@ -110,17 +110,22 @@ create_cluster() {
     printf "Log in to the OSD cluster using oc:\noc login --server=%s --username=kubeadmin --password=%s\n" "$(jq -r .api.url < "${CLUSTER_DETAILS_FILE}")" "$(jq -r .password < "${CLUSTER_CREDENTIALS_FILE}")"
 }
 
-install_rhmi() {
+install_addon() {
     local cluster_id
     local rhmi_name
     local infra_id
     local csv_name
+    local addon_id
+    local completion_phase
+
+    addon_id="${1}"
+    completion_phase="${2}"
 
     : "${USE_CLUSTER_STORAGE:=true}"
     cluster_id=$(get_cluster_id)
 
     echo "Applying RHMI Addon on a cluster with ID: ${cluster_id}"
-    echo '{"addon":{"id":"rhmi"}}' | ocm post "/api/clusters_mgmt/v1/clusters/${cluster_id}/addons"
+    echo "{\"addon\":{\"id\":\"${addon_id}\"}}" | ocm post "/api/clusters_mgmt/v1/clusters/${cluster_id}/addons"
 
     wait_for "oc --kubeconfig ${CLUSTER_KUBECONFIG_FILE} get rhmi -n ${RHMI_OPERATOR_NAMESPACE} | grep -q NAME" "rhmi installation CR to be created" "15m" "30"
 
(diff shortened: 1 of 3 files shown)


scripts/ocm/ocm.sh
```diff
@@ -156,6 +156,53 @@ install_rhmi() {
...
+
+    if [[ "${PATCH_CR_AWS_CM}" == true ]]; then
+        echo "Patching Cloud Resources AWS Strategies Config Map"
+        wait_for "oc --kubeconfig ${CLUSTER_KUBECONFIG_FILE} get configMap cloud-resources-aws-strategies -n ${RHMI_OPERATOR_NAMESPACE} | grep -q cloud-resources-aws-strategies" "cloud-resources-aws-strategies ready" "5m" "20"
+        oc --kubeconfig "${CLUSTER_KUBECONFIG_FILE}" patch configMap cloud-resources-aws-strategies -n "${RHMI_OPERATOR_NAMESPACE}" --type='json' -p '[{"op": "add", "path": "/data/_network", "value":"{ \"production\": { \"createStrategy\": { \"CidrBlock\": \"'10.1.0.0/23'\" } } }"}]'
+    fi
+
+    wait_for "oc --kubeconfig ${CLUSTER_KUBECONFIG_FILE} get rhmi ${rhmi_name} -n ${RHMI_OPERATOR_NAMESPACE} -o json | jq -r .status.stages.products.phase | grep -q completed" "rhmi installation" "90m" "300"
+    oc --kubeconfig "${CLUSTER_KUBECONFIG_FILE}" get rhmi "${rhmi_name}" -n ${RHMI_OPERATOR_NAMESPACE} -o json | jq -r '.status.stages'
+}
```
psturc, line 203: this function looks very similar to [`install_rhmi function`](https://github.com/integr8ly/delorean/blob/35f6431bf6408dc30c0d0bba04656df5a9fba190/scripts/ocm/ocm.sh#L113-L157) - do you think it would make sense to modify `install_rhmi` function and make it reusable for both type of installations?

sergioifg94: Agreed. I refactored them in a separate `install_addon` function that receives the addon id and completion phase as parameters

psturc:
LGTM, I'm going to verify this now

psturc:
@sergioifg94 installation of RHMI works fine, but when I try to install managed API, it gets stuck on creating secrets.

I think that is because in RHMI we were relying on Sendgrid service to create SMTP secret automatically after rhmi addon is enabled. And Sendgrid service doesn't work with Managed API addon yet.

Also recently we removed the requirement for smtp secret from RHMI operator, so it is no longer required to wait until SMTP secret is generated in order to start the installation. However in the script, there is [this check](https://github.com/integr8ly/delorean/blob/35f6431bf6408dc30c0d0bba04656df5a9fba190/scripts/ocm/ocm.sh#L311-L316) that verifies that smtp secret is present in rhmi-operator namespace.
I would probably update it to wait only for dms and pagerduty secrets.

psturc:
Maybe we should also update [the guide in this repo](https://github.com/integr8ly/delorean/blob/master/docs/ocm/README.md) as well, wdyt?

sergioifg94:
@psturc Thanks for letting me know. I updated the script to only wait for those 2 secrets, and updated that document to show how to install both RHMI and MGDAPI addons

PR #148, opened 2020-10-15
Openshift CI Release update for RHOAM

Boomatang:
These changes allow the use of a rhoam prefix to branch names.
Selectable by a input flag of --olmType. The default for the flag is integreatly-operator.
Images that are mirrored from prow has also been updated to suit the new install type.

**JIRA** 
https://issues.redhat.com/browse/MGDAPI-31
https://issues.redhat.com/browse/MGDAPI-245

**Requires https://github.com/integr8ly/integreatly-operator/pull/1155 to be merged**

Sample out put which would be created in a pr to openshift/release can been seen here. https://github.com/Boomatang/release/pull/4
Command used to create the pr was. values for the upstream was changed to from the openshift/release to Boomatang/release. this was a hard code value change and has been reverted in this pr.
```
GITHUB_TOKEN=<TOKEN> GITHUB_USER="Boomatang" ./delorean release openshift-ci-release --version 99.999.999 -o Boomatang --olmType managed-api-service
```

### DIFF
--- a/cmd/openshiftCIRelease.go
+++ b/cmd/openshiftCIRelease.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"context"
 	"fmt"
+	"github.com/integr8ly/delorean/pkg/types"
 	"os"
 	"os/exec"
 	"path"
@@ -302,11 +303,20 @@ func updateImageMirroringConfig(repoDir string, version *utils.RHMIVersion) erro
 		externalRegTemplate string
 	}
 
+	var operatorImage imageTemplate
+	operatorImage.internalRegTemplate = "%s/%s:integreatly-operator"
+
+	switch version.OlmType() {
+	case types.OlmTypeRhmi:
+		operatorImage.externalRegTemplate = "%s/integreatly-operator:%s"
+	case types.OlmTypeRhoam:
+		operatorImage.externalRegTemplate = "%s/managed-api-service:%s"
+	default:
+		operatorImage.externalRegTemplate = "%s/integreatly-operator:%s"
+	}
+
 	imageTemplates := []imageTemplate{
-		{
-			internalRegTemplate: "%s/%s:integreatly-operator",
-			externalRegTemplate: "%s/integreatly-operator:%s",
-		},
+		operatorImage,
 		{
 			internalRegTemplate: "%s/%s:integreatly-operator-test-harness",
 			externalRegTemplate: "%s/integreatly-operator-test-harness:%s",
@@ -389,7 +399,7 @@ func newOpenshiftCIReleaseCmd(f *openshiftCIReleaseCmdFlags) (*openshiftCIReleas
 		return nil, err
 	}
 	client := newGithubClient(token)
-	version, err := utils.NewRHMIVersion(releaseVersion)
+	version, err := utils.NewVersion(releaseVersion, olmType)
 	if err != nil {
 		return nil, err
 	}
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -20,6 +20,7 @@ var cfgFile string
 var integreatlyGHOrg string
 var integreatlyOperatorRepo string
 var releaseVersion string
+var olmType string
 
 var kubeconfigFile string
 
@@ -104,6 +105,7 @@ func init() {
 	releaseCmd.PersistentFlags().StringVarP(&integreatlyOperatorRepo, "repo", "r", DefaultIntegreatlyOperatorRepo, "Github repository")
 	releaseCmd.PersistentFlags().String("quayToken", "", fmt.Sprintf("Access token for quay. Can be set via the %s env var", strings.ToUpper(QuayTokenKey)))
 	viper.BindPFlag(QuayTokenKey, releaseCmd.PersistentFlags().Lookup("quayToken"))
+	releaseCmd.PersistentFlags().StringVarP(&olmType, "olmType", "", DefaultIntegreatlyOperatorRepo, "OLM type for the release. Valid inputs are \"integreatly-operator\" or \"managed-api-service\"")
 
 	defaultKubeconfigFilePath := ""
 	if home := homedir.HomeDir(); home != "" {
--- a/configurations/managed-tenants-addons-config-rhoam.yaml
+++ b/configurations/managed-tenants-addons-config-rhoam.yaml
@@ -1,0 +1,33 @@
+---
+addons:
+  - name: "rhoam-operator"
+    csv:
+      repo: "https://github.com/integr8ly/integreatly-operator.git"
+      path: "deploy/olm-catalog/managed-api-service"
+    channels:
+      - name: "stage"
+        directory: "rhoam-operator"
+        environment: "stage"
+        allow_pre_release: true
+      - name: "edge"
+        directory: "rhoam-operator-internal"
+        environment: "production"
+        allow_pre_release: false
+      - name: "stable"
+        directory: "rhoam-operator"
+        environment: "production"
+        allow_pre_release: false
+    override:
+      deployment:
+        name: "rhoam-operator"
+        container:
+          name: "rhoam-operator"
+          env_vars:
+            - name: "INSTALLATION_TYPE"
+              value: "managed-api"
+            - name: "USE_CLUSTER_STORAGE"
+              value: ""
+            - name: "ALERTING_EMAIL_ADDRESS"
+              value: "{{ alertingEmailAddress }}"
+
+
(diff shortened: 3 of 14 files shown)


cmd/openshiftCIRelease.go
```diff
@@ -389,15 +399,15 @@ func newOpenshiftCIReleaseCmd(f *openshiftCIReleaseCmdFlags) (*openshiftCIReleas
...
+	version, err := utils.NewRHMIVersion(releaseVersion, olmType)
 	if err != nil {
 		return nil, err
 	}
 	baseBranch := plumbing.NewBranchReferenceName(f.baseBranch)
 	return &openshiftCIReleaseCmd{
 		version:                 version,
 		baseBranch:              baseBranch,
-		releaseRepoInfoUpstream: &githubRepoInfo{owner: f.openshiftCIOrgUpstream, repo: f.openshiftCIRepo},
+		releaseRepoInfoUpstream: &githubRepoInfo{owner: "Boomatang", repo: f.openshiftCIRepo},
```
wei-lee, line 410: need to revert the hardcode value to use the flags

pkg/utils/rhmi_version.go
```diff
@@ -10,15 +11,16 @@ const releaseBranchNameTemplate = "prepare-for-release-%s"
...
+	build   string
+	major   string
+	minor   string
+	patch   string
+	olmType string
 }
 
 // NewRHMIVersion parse the integreatly version as a string and returns a Version object
-func NewRHMIVersion(version string) (*RHMIVersion, error) {
+func NewRHMIVersion(version string, olmType string) (*RHMIVersion, error) {
```
wei-lee, line 23: it looks like the only place where `olmType` used is for the `ReleaseBranchName` method? if this is the case, then it might be a much smaller change if we update the `ReleaseBranchName` method to take in the `olmType` as an extra optional parameter, rather than adding it as a parameter to the constructor? It doesn't feel right to have the `olmType` as part of a version.

pkg/utils/rhmi_version_test.go
```diff
@@ -25,6 +26,7 @@ func TestReleaseVersion(t *testing.T) {
 			expectError:    false,
 			isPatchRelease: false,
 			branchImageTag: "master",
+			olmType:        "integrealy-operator",
```
wei-lee, line 29: consider using the constants you have already defined.

wei-lee (changes requested):
Added a few comments. If this the only command that needs to be changed? Any other commands needs to be updated as well?

cmd/openshiftCIRelease.go
Boomatang: Will change

pkg/utils/rhmi_version.go
Boomatang: The initial work here was getting the Makefile for integreatly work for both rhmi and rhoam. Now that I have looked into the release pipeline there will be other functions that require the version to know about olm type. I am going to mark this pr as a WIP till I get the required updated to work with the pipeline in place. 
I understand the `olmType` seem wrong as part of the version. With the hindsight from looking into the release pipeline I feel putting it in the version will reduce the changes and check that would be required in other functions.

pkg/utils/rhmi_version_test.go
Boomatang: While change

Boomatang:
It does look like other commands will need to be changed.

pkg/utils/rhmi_version.go
wei-lee: How so? do you have an example?

Boomatang: An example, sure. As part of the release the help repo is been tagged the same as the release branch. This https://github.com/integr8ly/delorean/blob/master/cmd/tagRepository.go#L80-L81 gets called each time and we need to update the tag style to suit for `rhoam` Same thing happens with pushing images to quay. 

On the other side of this I do want to keep rhmi as the default so we don't have to go back updated existing pipelines for rhmi. Because of this I think do the version prefixing is easiest done in the RHMIversion stuct.

wei-lee: Ok, fair enough. In this case I would suggest that to keep the existing `NewRHMIVersion` method, which will by default continue using `integreaty-operator` as the olmType, and create a new method that will take the `olmType` parameter. This way you only need to update files that you actually need to override the default `olmType`, rather than changing them everywhere.

Boomatang: That is such a better idea. I never liked it passing the `olmType` while the method was called `NewRHMIVersion`. I will make the new method called `NewVersion`

cmd/root.go
```diff
@@ -104,6 +105,7 @@ func init() {
 	releaseCmd.PersistentFlags().StringVarP(&integreatlyOperatorRepo, "repo", "r", DefaultIntegreatlyOperatorRepo, "Github repository")
 	releaseCmd.PersistentFlags().String("quayToken", "", fmt.Sprintf("Access token for quay. Can be set via the %s env var", strings.ToUpper(QuayTokenKey)))
 	viper.BindPFlag(QuayTokenKey, releaseCmd.PersistentFlags().Lookup("quayToken"))
+	releaseCmd.PersistentFlags().StringVarP(&olmType, "olmType", "", DefaultIntegreatlyOperatorRepo, "OLM type for the release. Valid inputs are \"integreatly-operator\" or \"managed-api-service\"")
```
laurafitzgerald, line 108: I notice that we provide integreatly-operator as a default in the code and we specify here that valid input is either `integreatly-operator` or `managed-api-service`. I think this is mostly used by the pipeline but if someone runs the command with a typo in `managed-api-service` and it defaults to integreatly-operator is it just a case of running the command again correctly or does it handle a bad entry?

laurafitzgerald (commented):
Just taking a quick look. I think it's still in progress. Just one question inline. I also spotted that all the image mapping files that we use in openshift/release are using the versions without the v. e.g. prow mirrors versions 2.0->2.7 in quay but the images we use inside the manifest directories are in the format v2.1.0. I think that means that we don't actually use the images mirrored by prow? This seems historical and not related to your change here as that seems to be following the others, in the case of the mirroring mapping file at least.

cmd/root.go
Boomatang: It does not validated for a bad entry. If a service is put in wrongly the default is to act like a its `integreatly-operator` Are you suggesting that we error out if an unknown olm type is passed in.

philbrookes: I suppose the result of bad input is a PR to MT right? Which we could close? At the same time, maybe errorring out is more sensible here, could be confusing trying to work out where your release has gone, when there were no errors for your typo.

laurafitzgerald: Yeah it's definitely not a hill I'm willing to die on. I think if we have time to do it now it would be nice or alternatively we could push it out to a new JIRA so as not to block this work being finished in the time.

Boomatang: Ok I will get a possible fix for this into it today

philbrookes: yeah agreed, as we're pretty time sensitive on this PR, I'm happy to resolve this later and just get this merged as is for now.

philbrookes:
@Boomatang Can you rerun your test and confirm the output still looks OK? Can you also verify that the RHMI flow is unaffected?

Boomatang:
@philbrookes I reran my checks, found one spot which I was spelling integreatly wrong. That has now been forced pushed.

configurations/managed-tenants-addons-config-rhoam.yaml
```diff
@@ -0,0 +1,33 @@
...
+        name: "rhoam-operator"
+        container:
+          name: "rhoam-operator"
+          env_vars:
+            - name: "INSTALLATION_TYPE"
+              value: "managed-api"
+            - name: "USE_CLUSTER_STORAGE"
+              value: ""
+            - name: "ALERTING_EMAIL_ADDRESS"
+              value: "{{ alertingEmailAddress }}"
```
wei-lee, line 31: I am pretty sure you don't need to put this in a separate file. You can add it to the existing configuration file. When the command runs it will look up for the right configuration using name.

Boomatang: You are possible correct. Would it possible to get this merged and I will make a new PR to remove it. Its just the MR to ci-cd has already been merged which is using this config file. So that would also need to be updated. And we would like to use this in a release run today

wei-lee: sure

PR #151, opened 2020-10-26
MGDAPI-499 - add BU email and fix RHOAM env var overrides

matskiv:
JIRA: https://issues.redhat.com/browse/MGDAPI-499

Environment variable overrides were not getting applied, see https://gitlab.cee.redhat.com/service/managed-tenants/-/blob/c2a36cf1523bfb657ca1ede30f853e480868148d/addons/managed-api-service/bundles/0.2.0/managed-api-service.v0.2.0.clusterserviceversion.yaml.j2#L329
Also, BU email env var was missing.

I think that overrides didn't work because of the deployment and container names used.
I am changing those in the configuration file, but as an alternative, we could change them in the CSV in integreatly-operator repo. Wdyt @philbrookes @pwright @laurafitzgerald  ?

### DIFF
--- a/configurations/managed-tenants-addons-config-rhoam.yaml
+++ b/configurations/managed-tenants-addons-config-rhoam.yaml
@@ -19,9 +19,9 @@ addons:
         allow_pre_release: false
     override:
       deployment:
-        name: "rhoam-operator"
+        name: "rhmi-operator"
         container:
-          name: "rhoam-operator"
+          name: "rhmi-operator"
           env_vars:
             - name: "INSTALLATION_TYPE"
               value: "managed-api"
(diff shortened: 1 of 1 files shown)


matskiv:
@Boomatang you are familiar with this area now, would you mind reviewing, please?
Also, if you can suggest a way to test this, it would be great. I haven't tried running it with this change yet.

Boomatang:
familiar is one word for it. I don't what the best way of testing this. I have copied your changes to a different pr which is been merged sortly. There was an issue with the addon name also. https://github.com/integr8ly/delorean/pull/152

matskiv:
@Boomatang okay, can you please close the JIRA after your PR gets merged?
https://issues.redhat.com/browse/MGDAPI-499

PR #156, opened 2020-11-05
MGDAPI-419 Set env var for RHOAM deployment

CathalOConnorRH:
Update env var for RHOAM deployment

### DIFF
--- a/configurations/managed-tenants-addons-config-rhoam.yaml
+++ b/configurations/managed-tenants-addons-config-rhoam.yaml
@@ -19,9 +19,9 @@ addons:
         allow_pre_release: false
     override:
       deployment:
-        name: "rhmi-operator"
+        name: "rhoam-operator"
         container:
-          name: "rhmi-operator"
+          name: "rhoam-operator"
           env_vars:
             - name: "INSTALLATION_TYPE"
               value: "managed-api"


CathalOConnorRH:
@matskiv @Boomatang I think both of you know most about this change. Based on the changes we've made for https://issues.redhat.com/browse/MGDAPI-419 and the rhoam pipeline release script.

CathalOConnorRH:
@laurafitzgerald Would you be able to review this also please?

laurafitzgerald:
From a quick look it seems fine. I noticed that we do a check on the csv for the deployment and container [here](https://github.com/integr8ly/delorean/blob/3902a19ab27164e7a16925b2f62d46b5a83ff251/cmd/osdAddonRelease_test.go#L306-L313) If this is used in the generation of the csv then that might also need to be updated to account for both RHMI and RHOAM releases. @Boomatang or @MStokluska might be able to give some insight into whether this would be the case or not.

matskiv:
AFAIK deployment and container names here should match what's in [deploy/operator.yaml](https://github.com/integr8ly/integreatly-operator/blob/master/deploy/operator.yaml#L4-L20), at the moment it's still `rhmi-operator`.
and it`s not being changed in related PR - https://github.com/integr8ly/integreatly-operator/pull/1344/files

Is it changed by some release script on the fly, or does the deployment/container name stays `rhmi-operator`

CathalOConnorRH:
it's being changed by the prepare-release.sh script 
https://github.com/integr8ly/integreatly-operator/blob/master/scripts/prepare-release.sh#L117:L128

matskiv:
Cool, thanks @CathalOConnorRH 
/lgtm
/approve

matskiv:
Oops, I forgot about Laura's comment and rushed with the approval, sorry.
Good that I don't have permissions for this. 😂

Boomatang:
On @laurafitzgerald comment that test case will use RHMI and the configure files for it. I don't see that been an issue in this case.

laurafitzgerald:
@Boomatang thanks for the detail @CathalOConnorRH you good with this one now? Is there any other review/verification required here or are we good to approve?

CathalOConnorRH:
@laurafitzgerald Im happy with this if you are. I think we should be ok to approve

PR #157, opened 2020-11-05
feat: option to specify number of cluster nodes

psturc:
### How to verify

```bash
make ocm/cluster.json | jq -r .nodes.compute
# You should get 4 on the output (default number of nodes for single-AZ cluster)

make ocm/cluster.json MULTI_AZ=true | jq -r .nodes.compute
# You should get 9 on the output (default number of nodes for multi-AZ cluster)

make ocm/cluster.json COMPUTE_NODES_COUNT=123 | jq -r .nodes.compute
# You should get 123 on the output (custom number of nodes). Same for multi AZ cluster:

make ocm/cluster.json MULTI_AZ=true COMPUTE_NODES_COUNT=123 | jq -r .nodes.compute
```

### DIFF
--- a/scripts/ocm/ocm.sh
+++ b/scripts/ocm/ocm.sh
@@ -60,6 +60,7 @@ create_cluster_configuration_file() {
     : "${OPENSHIFT_VERSION:=}"
     : "${PRIVATE:=false}"
     : "${MULTI_AZ:=false}"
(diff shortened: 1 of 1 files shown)


CathalOConnorRH:
tested script locally, works as described. 
/lgtm

PR #160, opened 2020-11-17
Update test-containers.yaml

wei-lee:


wei-lee:
@mikenairn @StevenTobin can you approve the PR? Want to see if I can force a rebuild of the cli image to see if the cert issue is now fixed.

PR #173, opened 2021-01-15
mgdapi-304-create-manifest

MStokluska:
Changes related to https://issues.redhat.com/browse/MGDAPI-304

Do not merge - manifest generation script must be updated first

To verify changes:
Run from this repo
- make build/cli
- update the "DefaultIntegreatlyGithubOrg" to your github org in root.go file
- make sure that your local branch of integreatly operator is up to date
- generate new client access token
- export GITHUB_TOKEN=<YOUR ACCESS TOKEN>
- export GITHUB_USER=<YOUR USER NAME> 
- ./delorean release create-rhoam-manifest --version=1.1.0 --olmType=managed-api-service --branch=master

Verify that a new PR has been submitted against your branch with name `rhoam-manifest-for-release-rhoam-1.1.0`
NOTE: updates to the actual script call might be made later on, once scripts in the integreatly repo got updated.

### DIFF
--- a/cmd/createProdsecManifest.go
+++ b/cmd/createProdsecManifest.go
@@ -1,0 +1,265 @@
+package cmd
+
+import (
+	"context"
+	"fmt"
+	"github.com/go-git/go-git/v5"
+	"github.com/go-git/go-git/v5/plumbing"
+	"github.com/go-git/go-git/v5/plumbing/object"
+	"github.com/go-git/go-git/v5/plumbing/transport/http"
+	"github.com/google/go-github/v30/github"
+	"github.com/integr8ly/delorean/pkg/services"
+	"github.com/integr8ly/delorean/pkg/utils"
+	"github.com/spf13/cobra"
+	"io"
+	"os"
+	"os/exec"
+	"path"
+	"time"
+)
+
+type createProdsecManifestCmdFlags struct {
+	baseBranch     string
+	manifestScript string
+}
+
+type createProdsecManifestCmd struct {
+	version         *utils.RHMIVersion
+	repoInfo        *githubRepoInfo
+	baseBranch      plumbing.ReferenceName
+	githubPRService services.PullRequestsService
+	manifestScript  string
+	gitUser         string
+	gitPass         string
+	gitCloneService services.GitCloneService
+	gitPushService  services.GitPushService
+}
+
+func init() {
+	f := &createProdsecManifestCmdFlags{}
+
+	cmd := &cobra.Command{
+		Use:   "create-prodsec-manifest",
+		Short: "Create a production manifest of a given version and olm type",
+		Run: func(cmd *cobra.Command, args []string) {
+			c, err := newCreateProdsecManifestCmd(f)
+			if err != nil {
+				handleError(err)
+			}
+			var repoDir string
+			if repoDir, err = c.run(cmd.Context()); err != nil {
+				handleError(err)
+			}
+			if repoDir != "" {
+				fmt.Println("Remove temporary directory:", repoDir)
+				if err = os.RemoveAll(repoDir); err != nil {
+					handleError(err)
+				}
+			}
+		},
+	}
+
+	releaseCmd.AddCommand(cmd)
+	cmd.Flags().StringVarP(&f.baseBranch, "branch", "b", "master", "Base branch of the manifest generation")
+	cmd.Flags().StringVar(&f.manifestScript, "manifestGenerationScript", "scripts/prodsec-manifest-generator.sh", "Relative path to the script to run before creating the PR")
+}
+
+func newCreateProdsecManifestCmd(f *createProdsecManifestCmdFlags) (*createProdsecManifestCmd, error) {
+	var token string
+	var err error
+	if token, err = requireValue(GithubTokenKey); err != nil {
+		return nil, err
+	}
+	var user string
+	if user, err = requireValue(GithubUserKey); err != nil {
+		return nil, err
+	}
+	client := newGithubClient(token)
+	repoInfo := &githubRepoInfo{owner: integreatlyGHOrg, repo: integreatlyOperatorRepo}
+	baseBranch := plumbing.NewBranchReferenceName(f.baseBranch)
+	version, err := utils.NewVersion(releaseVersion, olmType)
+	if err != nil {
+		return nil, err
+	}
+	return &createProdsecManifestCmd{
+		version:         version,
+		repoInfo:        repoInfo,
+		baseBranch:      baseBranch,
+		manifestScript:  f.manifestScript,
+		githubPRService: client.PullRequests,
+		gitUser:         user,
+		gitPass:         token,
+		gitCloneService: &services.DefaultGitCloneService{},
+		gitPushService:  &services.DefaultGitPushService{},
+	}, nil
+}
+
+func (c *createProdsecManifestCmd) run(ctx context.Context) (string, error) {
+	// Clone the repo
+	fmt.Println(fmt.Sprintf("Clone repo from %s/%s/%s.git to a temporary directory", githubURL, c.repoInfo.owner, c.repoInfo.repo))
+	repoDir, gitRepo, err := c.gitCloneService.CloneToTmpDir("integreatly-operator", fmt.Sprintf("%s/%s/%s.git", githubURL, c.repoInfo.owner, c.repoInfo.repo), c.baseBranch)
+	if err != nil {
+		return "", err
+	}
+
+	fmt.Println("Repo cloned to", repoDir)
+	gitRepoTree, err := gitRepo.Worktree()
+	if err != nil {
+		return "", err
+	}
+
+	// Checking out the OLM_TYPE-release-VERSION branch as the manifest generation script must be run from this branch
+	releasedBranchName := c.version.ReleaseBranchName()
+	fmt.Println(fmt.Sprintf("Checkout branch %s", releasedBranchName))
+	if err = checkoutBranch(gitRepoTree, false, false, releasedBranchName); err != nil {
+		return "", err
+	}
+
+	// Invoking manifest generation script
+	fmt.Println(fmt.Sprintf("Generate manifest: %s", c.manifestScript))
+	if err = c.runManifestScript(repoDir); err != nil {
+		return "", err
+	}
+
+	manifestFileName := fmt.Sprintf("%s-production-release-manifest.txt", c.version.NameByOlmType())
+	manifestPath := fmt.Sprintf("%s/prodsec-manifests/%s", repoDir, manifestFileName)
+	temporaryManifestPath := fmt.Sprintf("%s/../temporary-manifest.txt", repoDir)
+
+	// Saving generated manifest file - reason for it is as we need to checkout master without saving local changes to the repo.
+	if err = copyManifestFile(manifestPath, temporaryManifestPath); err != nil {
+		return "", err
+	}
+
+	// Checking out master branch to be able to checkout new branch from it
+	fmt.Println("Checkout master branch")
+	if err = checkoutBranch(gitRepoTree, true, false, "master"); err != nil {
+		return "", err
+	}
+
+	// Preparing new branch called OLM_TYPE-manifest-for-release-OLM_TYPE-VERSION of release
+	manifestReleaseBranchName := c.version.PrepareProdsecManifestBranchName()
+	fmt.Println(fmt.Sprintf("Create new branch %s", manifestReleaseBranchName))
+	if err = checkoutBranch(gitRepoTree, true, true, manifestReleaseBranchName); err != nil {
+		return "", err
+	}
+
+	// Restoring saved file to a new branch
+	if err = copyManifestFile(temporaryManifestPath, manifestPath); err != nil {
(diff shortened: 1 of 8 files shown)


cmd/createRhoamManifest.go
```diff
@@ -0,0 +1,269 @@
...
+	}
+
+	// Invoking manifest generation script
+	fmt.Println(fmt.Sprintf("Generate manifest: %s", c.manifestScript))
+	if err = c.runManifestScript(repoDir); err != nil {
+		return "", err
+	}
+
+	manifestFileName := "rhoam-production-release-manifest.txt"
+	manifestPath := fmt.Sprintf("%s/rhoam-manifests/%s", repoDir, manifestFileName)
```
Boomatang, line 125: If my first comment on the command rename is seen a valuable, these string would possible need to be created based on the olmType. Putting this comment as a marker.

pkg/utils/rhmi_version.go
```diff
@@ -148,6 +149,10 @@ func (v *RHMIVersion) PrepareReleaseBranchName() string {
 	return fmt.Sprintf(releaseBranchNameTemplate, v.TagName())
 }
 
+func (v *RHMIVersion) PrepareRhoamManifestBranchName() string {
```
Boomatang, line 152: This function should be more generic and switch the output depending on the olmType.

cmd/createRhoamManifest.go
```diff
@@ -0,0 +1,269 @@
...
+		pr, _, err = c.githubPRService.Create(ctx, c.repoInfo.owner, c.repoInfo.repo, req)
+		if err != nil {
+			return err
+		}
+	}
+	fmt.Println(fmt.Sprintf("PR created: %s", pr.GetHTMLURL()))
+	return nil
+}
+
+func checkoutBranch(repoWorkTree *git.Worktree, keepChanges bool, create bool, branchName string) error {
```
Boomatang, line 231: To me I feel this function may have been created before and if not should this be moved to pkg/utils

cmd/createRhoamManifest_test.go
```diff
@@ -0,0 +1,162 @@
...
+
+func TestCreateRhoamManifestRelease(t *testing.T) {
+	cases := []struct {
+		description string
+		cmd         func() *createRhoamManifestCmd
+		expectError bool
+		verify      func(repoDir string) error
+	}{
+		{
+			description: "should successfully generate the correct manifest",
```
Boomatang, line 67: Do we need to check the generate of a manifest which has not changed

Boomatang (commented):
I know there is some changes coming so this is the first pass on it. This is only the code review I have not done any other steps. There maybe more need in the units but I will check that when I start running this local.

cmd/createRhoamManifest.go
```diff
@@ -0,0 +1,269 @@
...
+	gitCloneService services.GitCloneService
+	gitPushService  services.GitPushService
+}
+
+func init() {
+	f := &createRhoamManifestCmdFlags{}
+
+	cmd := &cobra.Command{
+		Use:   "create-rhoam-manifest",
+		Short: "Create a rhoam manifest of a given version",
```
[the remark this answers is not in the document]
MStokluska: this can change into `create-manifest`, however I think that `create-manifest` is too broad. As for the difference between rhmi and rhoam, reason why it is not checked is because it makes no difference. Go mod graph is called on the entire repo, not only rhoam related packages. If we ever decide to have a different manifest for rhmi production, all we would do is edit the script itself

cmd/createRhoamManifest.go
```diff
@@ -0,0 +1,269 @@
...
+				fmt.Println("Remove temporary directory:", repoDir)
+				if err = os.RemoveAll(repoDir); err != nil {
+					handleError(err)
+				}
+			}
+		},
+	}
+
+	releaseCmd.AddCommand(cmd)
+	cmd.Flags().StringVarP(&f.baseBranch, "branch", "b", "master", "Base branch of the manifest generation")
```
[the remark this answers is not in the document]
MStokluska: "master" and the baseBranch from jenkins pipeline build, which is also called `baseBranch`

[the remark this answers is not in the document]
Boomatang: If this is the name that is coming from jenkins I am happy. To me it seems a bit vague of a description. I am right in saying this will be the branch any PR's would be create against when running the command.

cmd/createProdsecManifest.go
```diff
@@ -0,0 +1,273 @@
...
+
+	return repoDir, nil
+}
+
+func (c *createProdsecManifestCmd) runManifestScript(repoDir string) error {
+	if err := os.Chmod(path.Join(repoDir, c.manifestScript), 0755); err != nil {
+		return err
+	}
+
+	envs := []string{fmt.Sprintf("OLM_TYPE=%s", c.version.OlmType()), "TYPE_OF_MANIFEST=production", fmt.Sprintf("PATH=%s", os.Getenv("PATH")), "HOME=/tmp"}
```
Boomatang, line 177: `"TYPE_OF_MANIFEST=production"`, will there ever be a need for this to be anything else?

cmd/createProdsecManifest.go
```diff
@@ -0,0 +1,273 @@
...
+}
+
+func (c *createProdsecManifestCmd) createPRIfNotExists(ctx context.Context, releaseBranchName string) error {
+	h := fmt.Sprintf("%s:%s", c.repoInfo.owner, releaseBranchName)
+	prOpts := &github.PullRequestListOptions{Base: c.baseBranch.String(), Head: h}
+	pr, err := findPRForRelease(ctx, c.githubPRService, c.repoInfo, prOpts)
+	if err != nil && !isManifestPRNotFoundError(err) {
+		return err
+	}
+	if pr == nil {
```
Boomatang, line 217: Can we have a chat about this logic, I think we can improve the readability here a bit.

cmd/createProdsecManifest.go
```diff
@@ -0,0 +1,273 @@
...
+		}
+	}
+	fmt.Println(fmt.Sprintf("PR created: %s", pr.GetHTMLURL()))
+	return nil
+}
+
+func checkoutBranch(repoWorkTree *git.Worktree, keepChanges bool, create bool, branchName string) error {
+	branch := plumbing.NewBranchReferenceName(branchName)
+	if err := repoWorkTree.Checkout(&git.CheckoutOptions{
+		Force:  keepChanges,
```
Boomatang, line 238: During the checkout of a branch what changes are you expected need to be kept? I don't think we need the `Force: keepChanges,`

cmd/createProdsecManifest_test.go
```diff
@@ -0,0 +1,199 @@
...
+
+				content, err := ioutil.ReadFile(path.Join(repoDir, "prodsec-manifests/rhoam-production-release-manifest.txt"))
+				if err != nil {
+					return err
+				}
+				if strings.Index(string(content), "dummy manifest dependency") < 0 {
+					return fmt.Errorf("expected: %s, but got %s", "dummy manifest dependency", string(content))
+				}
+				return nil
+			},
```
Boomatang, line 130: The verify function is the same in both cases expect the path to the manifest file. Is there anything we can do to tidy this up and not repeat code.

cmd/createProdsecManifest_test.go
```diff
@@ -0,0 +1,199 @@
...
+		gitUser:         "testuser",
+		gitPass:         "testpass",
+		gitCloneService: cloneService,
+		gitPushService:  pushService,
+		githubPRService: prService,
+	}
+}
+
+func TestCreateRhoamManifestRelease(t *testing.T) {
+	cases := []struct {
```
Boomatang, line 60: I would like to see a test that checks what happens if manifest does not need updating and the PR is tried to be updated.

pkg/utils/rhmi_version.go
```diff
@@ -148,6 +152,17 @@ func (v *RHMIVersion) PrepareReleaseBranchName() string {
 	return fmt.Sprintf(releaseBranchNameTemplate, v.TagName())
 }
 
+func (v *RHMIVersion) PrepareProdsecManifestBranchName() string {
```
Boomatang, line 155: Can we add some unit tests for this function

pkg/utils/rhmi_version.go
```diff
@@ -165,3 +180,14 @@ func (v *RHMIVersion) ReleaseBranchImageTag() string {
 func (v *RHMIVersion) OlmType() string {
 	return v.olmType
 }
+
+func (v *RHMIVersion) NameByOlmType() string {
```
Boomatang, line 184: Can we add some unit tests for this function

Boomatang (changes requested):
Mostly looks good, I do have a few question, comments and requests

cmd/createProdsecManifest.go
MStokluska: no, it's for production only.

cmd/createProdsecManifest.go
```diff
@@ -0,0 +1,273 @@
...
+	gitCloneService services.GitCloneService
+	gitPushService  services.GitPushService
+}
+
+func init() {
+	f := &createProdsecManifestCmdFlags{}
+
+	cmd := &cobra.Command{
+		Use:   "create-prodsec-manifest",
+		Short: "Create a production manifest of a given version and olm type",
```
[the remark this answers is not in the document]
MStokluska: It's the version you are aiming to promote to production

cmd/createProdsecManifest.go
MStokluska: None really, it doesn't work exactly as if you would locally checkout branches. For example, if you checkout from release-1.1.0 to master, all differences are kept and master branch gets populated with release-1.1.0 data. Force: set to true forces any changes between branches to be excluded from branching.

cmd/createProdsecManifest_test.go
MStokluska: Not that I'm aware as they are cases, if you know of a way how could we move it into 1 case please let me know, although I think a case - should be unique with values hard coded to ensure that the particular case covers particular settings.

pkg/utils/rhmi_version.go
MStokluska: Do you think it is necessary as it's only a switch statement? Following the approach used so far, not creating unit test cases for simple functions that reside in rhmi_version.go I decided to skip it too.

pkg/utils/rhmi_version.go
MStokluska: Same as above, I think we are ok to skip unit tests of function that contains only switch statement with default value in worst case scenario.

pkg/utils/rhmi_version.go
Boomatang: If the function is simple it should be simple create the unit test. Delorean has a good unit test habit, we should try to maintain that good practice as best we can. And the fact it's in the rhmi_version.go file maybe a bigger reason to have the tests. Nearly every command uses the code from this file.

pkg/utils/rhmi_version.go
Boomatang: Same answer as above, it's about keeping the quaintly high for the next person making changes.

cmd/createProdsecManifest.go
Boomatang: Ok I am happy with answer (I think) it seems a bit backwards

cmd/createProdsecManifest_test.go
Boomatang: It is the function been create inline on the verify step which can be refactored to be shared between both. Not the test case its self. I am happy with the two cases, would even like to see more.

pkg/utils/rhmi_version.go
MStokluska: thanks, updated!

pkg/utils/rhmi_version.go
MStokluska: thanks, updated!

cmd/createProdsecManifest.go
MStokluska: Resolving as per our conversation, so our main points out of this, delorean cli could use a slight refactor to 3 existing cli commands that use same functions, these functions should be made in utils to become shared functions. Also, as per our conversation, updating some unit tests would be useful. Thanks @Boomatang

cmd/createProdsecManifest.go
MStokluska: Updated the "keepChanges" to be more of a reflection of what we are doing here.

cmd/createProdsecManifest_test.go
MStokluska: Good spot, thanks for helping me out with it.

cmd/createProdsecManifest_test.go
MStokluska: As per our conversation, we have agreed that in line what's done in other functions that use the similar function to create a release branch, this could be part in creating this function as shared in utils and then the appropriate test cases could be done. Thanks.

MStokluska:
Thanks for the review @Boomatang, we have run the verification against my own integreatly repo and it works as expected.

PR #176, opened 2021-01-29
use http to verify image tag - remove use of docker

briangallagher:
#### Build:
`make build/cli`

#### Success:
```
./delorean ews get-image-origin --imageType rateLimiting --imageTag v1.4.0
docker.io/envoyproxy/ratelimit:v1.4.0
```

#### Fail:
```
./delorean ews get-image-origin --imageType rateLimiting --imageTag v1.4.1
Image tag does not exist: https://index.docker.io/v1/repositories/envoyproxy/ratelimit/tags/v1.4.1 http Status 404 NOT FOUND
```


CathalOConnorRH:
verified, 

/lgtm

pkg/utils/image_replace.go
```diff
@@ -61,19 +59,30 @@ var ImageSubs = map[string]*ImageDetails{
...
 	if err != nil {
-		fmt.Println("Error pulling image: ", err)
+		fmt.Println("Error making http call: ", err)
 		return "", err
 	}
-	return image, nil
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		fmt.Println(fmt.Sprintf("Image tag does not exist: %s http Status %s", url, resp.Status))
+		return "", err
```
laurafitzgerald, line 82: I don't think there will be an err here. err from http.Client.Do needs to be nil to get to this code. I think you would need to generate a new error here?

briangallagher: yep, good catch

pkg/utils/image_replace.go
```diff
@@ -61,19 +59,29 @@ var ImageSubs = map[string]*ImageDetails{
...
+
+	// Verify that the image exists in image repo
+
+	httpClient := &http.Client{}
+	url := "https://index.docker.io/v1/repositories/envoyproxy/ratelimit/tags/" + imageTag
+
+	req, err := http.NewRequest("GET", url, nil)
 	if err != nil {
-		fmt.Println("Error getting docker client: ", err)
+		fmt.Println("Error getting request: ", err)
```
laurafitzgerald, line 70: Minor nit. do we use fmt.Println generally in this repo? I think it could be cleaner to do an error.Wrap on the return rather than fmt.Print?

briangallagher: Yeah, it's used throughout. Probably a separate JIRA to handle logs and errors cleaner throughout the code base. wdyt?

PR #180, opened 2021-02-16
updating network configuration to match the defaults set in ocm

laurafitzgerald:
**What**
$title

**Why**
ocm uses these defaults
rosa also uses these defaults

I'm not sure about the host prefix. Looks like the default in ocm is `/23` but I see clusters with `23` and it's working as well. It's not clear to me what the difference is so that change can be removed if necessary.

### DIFF
--- a/templates/ocm/cluster-template.json
+++ b/templates/ocm/cluster-template.json
@@ -7,10 +7,10 @@
     "multi_az": false,
     "name": "my_cluster_name",
     "network": {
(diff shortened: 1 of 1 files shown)


laurafitzgerald:
/hold
Discussing this one in this JIRA -> https://issues.redhat.com/browse/MGDAPI-1274

laurafitzgerald:
Gonna close this one for now. It's liked in the related JIRA.

PR #187, opened 2021-03-05
[MGDAPI-1349] Updated prow config version

makslion:
## What 
updated what config file is used to generate prow checks 

## Why 
https://issues.redhat.com/browse/MGDAPI-1349
To include checks for files generated by the kubebuilder


makslion:
@laurafitzgerald boop

laurafitzgerald:
/approve
cc @Boomatang @MStokluska for visibility and `/lgtm`
context is we want this check to be in place for any new release branch config but it's only present for v1.3 onwards as that was when the operator-sdk change landed.

Boomatang:
@laurafitzgerald If I understand your comment correctly. We will not be updating that file version number with every release.

PR #189, opened 2021-03-16
Feat: Mark Aborted stages as skipped test

Boomatang:
When exported to junit xml the aborted stages where not marked as skipped or failed. By marking the aborted as skipped we can get some more insight into the pipelines using tools like report portal.

Format can be checked by building the cli and running the following command
`./delorean pipeline junit-report --input ./pkg/utils/testdata/jenkins/aborted-pipeline-status.json --output /tmp/pipeline.xml`

### DIFF
--- a/pkg/utils/jenkins.go
+++ b/pkg/utils/jenkins.go
@@ -91,8 +91,9 @@ type PipelineRun struct {
 }
 
 const (
-	TestSuiteName           = "pipeline-status"
-	PipelineRunStatusFailed = "FAILED"
+	TestSuiteName            = "pipeline-status"
+	PipelineRunStatusFailed  = "FAILED"
+	PipelineRunStatusAborted = "ABORTED"
 )
 
 // ToJUnitSuites will convert the status of the pipeline run into JUnit test suites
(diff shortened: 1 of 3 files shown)


pkg/utils/jenkins.go
```diff
@@ -114,12 +115,18 @@ func (p *PipelineRun) ToJUnitSuites() (*JUnitTestSuites, error) {
...
-		if s.Status == PipelineRunStatusFailed {
+
+		switch s.Status {
+		case PipelineRunStatusFailed:
 			ts.Failures++
 			tc.Failure = &JUnitFailure{
 				Message: s.Error.Message,
 				Type:    s.Error.Type,
 			}
+		case PipelineRunStatusAborted:
```
psturc, line 126: Not sure if marking all `Aborted` stages as `Skipped` makes sense.

Usually, when pipeline **fails** at some stage, it marks that stage + all subsequent stages as `FAILED`.
When pipeline is **aborted**, it means that it was manually aborted by user or system (e.g due to timeout). Similarly as for pipeline failure, it marks the aborted stage + all subsequent stages, but as `ABORTED`.

What I would propose is

**For FAILED pipeline:**
* Mark only the first failed stage as `Failed`
* subsequent stages should be marked as `Skipped`

**For ABORTED pipeline:**
* Mark only the first aborted stage as `Failed`
* subsequent stages should be marked as `Skipped`

Ideally we should also distinguish between "aborted by a user" and "aborted by system", but because we're currently tracking nightly pipelines in Reportportal, it is unlikely that the pipeline is aborted by a user...

Wdyt?

Boomatang:
@psturc I have push the update that marks the first `failed` or `aborted` stage as a failure and the rest as skipped. I did how ever set it so the first `failed` stage gets record as failed even if there was an aborted stage before hand. Same applies the `aborted` stages.

pkg/utils/jenkins.go
psturc: @Boomatang hmm I think it is possible, but we probably don't have that logic implemented in our nightly pipelines (I'm not aware of it)

Boomatang: I guess it's something that we need to keep an eye out for and maybe do something about it if we ever come across that scenario.

PR #195, opened 2021-04-13
MGDAPI-1309

psturc:
### JIRA
https://issues.redhat.com/browse/MGDAPI-1309

### DIFF
--- a/pkg/utils/jenkins.go
+++ b/pkg/utils/jenkins.go
@@ -5,6 +5,7 @@ import (
 	"encoding/xml"
 	"fmt"
 	"io"
+	"strings"
 	"time"
 )
 
@@ -91,13 +92,14 @@ type PipelineRun struct {
 }
 
 const (
-	TestSuiteName            = "pipeline-status"
-	PipelineRunStatusFailed  = "FAILED"
-	PipelineRunStatusAborted = "ABORTED"
+	TestSuiteName             = "pipeline-status"
+	PipelineRunStatusFailed   = "FAILED"
+	PipelineRunStatusAborted  = "ABORTED"
+	PipelineRunStatusUnstable = "UNSTABLE"
 )
 
 // ToJUnitSuites will convert the status of the pipeline run into JUnit test suites
-func (p *PipelineRun) ToJUnitSuites() (*JUnitTestSuites, error) {
+func (p *PipelineRun) ToJUnitSuites(filter string) (*JUnitTestSuites, error) {
 	suites := &JUnitTestSuites{
 		Suites: []JUnitTestSuite{},
 	}
@@ -145,6 +147,18 @@ func (p *PipelineRun) ToJUnitSuites() (*JUnitTestSuites, error) {
 					Message: s.Error.Message,
 				}
 			}
+		case PipelineRunStatusUnstable:
+			ts.Failures++
+			tc.Failure = &JUnitFailure{
+				Message: s.Error.Message,
+				Type:    s.Error.Type,
+			}
+		}
+
+		// If creating a report only from stages with a specific prefix,
+		// skip adding stage names to the report that do not match the prefix
+		if filter != "" && !strings.Contains(s.Name, filter) {
+			continue
 		}
 
 		ts.TestCases = append(ts.TestCases, tc)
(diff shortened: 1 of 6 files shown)


scripts/ocm/ocm.sh
```diff
@@ -179,8 +180,10 @@ install_addon() {
         oc --kubeconfig "${CLUSTER_KUBECONFIG_FILE}" patch configMap cloud-resources-aws-strategies -n "${OPERATOR_NAMESPACE}" --type='json' -p '[{"op": "add", "path": "/data/_network", "value":"{ \"production\": { \"createStrategy\": { \"CidrBlock\": \"'10.1.0.0/23'\" } } }"}]'
     fi
 
-    wait_for "oc --kubeconfig ${CLUSTER_KUBECONFIG_FILE} get rhmi ${rhmi_name} -n ${OPERATOR_NAMESPACE} -o json | jq -r ${completion_phase} | grep -q completed" "rhmi installation" "90m" "300"
-    oc --kubeconfig "${CLUSTER_KUBECONFIG_FILE}" get rhmi "${rhmi_name}" -n "${OPERATOR_NAMESPACE}" -o json | jq -r '.status.stages'
+    if [[ "${WAIT}" == true ]]; then
+        wait_for "oc --kubeconfig ${CLUSTER_KUBECONFIG_FILE} get rhmi ${rhmi_name} -n ${OPERATOR_NAMESPACE} -o json | jq -r ${completion_phase} | grep -q completed" "rhmi installation" "90m" "300"
+        oc --kubeconfig "${CLUSTER_KUBECONFIG_FILE}" get rhmi "${rhmi_name}" -n "${OPERATOR_NAMESPACE}" -o json | jq -r '.status.stages'
+    fi
```
Boomatang, line 186: I am worried this is changing the default behaviour and how it might affect existing pipelines. Should this set the default to wait and if the we pass a no wait flag to stop the install. I'm not sure what do you think? Personal I am not a fan of changing the default behaviour.

Boomatang (commented):
I am mostly happy with the changes. There is a few points that I don't not follow understand and have asked the questions in the comments.

scripts/ocm/ocm.sh
psturc: it is set to wait by default [here](0716e58baec8c2a636cfd8c12b40bc199ff6c16653892R141)

pkg/utils/jenkins.go
```diff
@@ -145,6 +147,18 @@ func (p *PipelineRun) ToJUnitSuites() (*JUnitTestSuites, error) {
 					Message: s.Error.Message,
 				}
 			}
+		case PipelineRunStatusUnstable:
+			ts.Failures++
+			tc.Failure = &JUnitFailure{
+				Message: s.Error.Message,
+				Type:    s.Error.Type,
+			}
+		}
```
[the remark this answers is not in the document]
psturc: I thought marking the "UNSTABLE" stage as failed makes sense, at least in case of nightly pipelines.

In "functional tests stage" in nightly pipeline, we don't allow it to be marked as failed, because Jenkins would automatically mark the **whole build** as FAILED, which might be misleading.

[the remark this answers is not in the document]
Boomatang: Ya I can see your point. I'm guessing it would skew it in reportporatl a small bit but I guess having it as a failure is far better than not doing any thing about at all. I am happy with this.

scripts/ocm/ocm.sh
Boomatang: Ok, if I am correct in saying this does not affect the existing pipelines I am happy.

scripts/ocm/ocm.sh
```diff
@@ -380,6 +383,7 @@ Optional exported variables:
 - USE_CLUSTER_STORAGE               true/false - use OpenShift/AWS storage (default: true)
 - ALERTING_EMAIL_ADDRESS            email address for receiving alert notifications
 - SELF_SIGNED_CERTS                 true/false - cluster certificate can be invalid
+- WAIT                              true/false - wait for install to complete
```
Boomatang, line 386: As this defaults to `true`, we should note that in this helper text.

Boomatang (commented):
Happy with the answer to my questions. There is one one change I would like as a result of the answer.

scripts/ocm/ocm.sh
psturc: good point, fixed

PR #204, opened 2021-05-18
breaking: reportportal v4 -> v5 updates

psturc:
https://issues.redhat.com/browse/MGDAPI-1752

### DIFF
--- a/cmd/reportPortalImport.go
+++ b/cmd/reportPortalImport.go
@@ -48,7 +48,7 @@ type reportPortalImportCmd struct {
 
 type reportProcessResult struct {
 	s3ObjKey   string
-	rpLaunchId string
+	rpLaunchId int
 }
 
 type testMetadata struct {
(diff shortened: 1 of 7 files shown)


Boomatang:
There are unit tests that need addressing

cmd/reportPortalImport.go
```diff
@@ -190,16 +190,23 @@ func (c *reportPortalImportCmd) processReportFile(ctx context.Context, object *s
 	if err != nil {
 		return nil, err
 	}
-	fmt.Println(fmt.Sprintf("[%s] File uploaded. Update the launch to add tags with id %s", *object.Key, importResp.GetLaunchId()))
+	fmt.Println(fmt.Sprintf("[%s] File uploaded. Get the Launch Id for Launch UUID %s", *object.Key, importResp.GetLaunchUuid()))
```
psturc, line 193: reportportal API in v5 now returns launch UUID instead of id.

so now in order to update the launch after import we need to get its ID first, which is what `c.rpLaunchService.Get` method is for

PR #208, opened 2021-05-31
Remove marking the unstable stages as failures.

Boomatang:
This is to address the extra work in ReportPortal caused by adding the extra failure marker.


MStokluska:
Verified in unit tests, the UNSTABLE stage is no longer marked as failed.
The only question was about the necessity of adding the `expectedTests` int, but we have agreed with Jim that it is good to keep it there for readibility.
/lgmt

PR #210, opened 2021-06-01
MGDAPI-1633: add param addon-resource-required to addon_payload

austincunningham:
# what
add param addon-resource-required to addon_payload

### DIFF
--- a/scripts/ocm/ocm.sh
+++ b/scripts/ocm/ocm.sh
@@ -157,7 +157,7 @@ install_addon() {
 
     # Add mandatory "cidr-range" and "addon-managed-api-service" (quota) params with default value in case of rhoam (managed-api-service) addon 
     if [[ "${addon_id}" == "managed-api-service" ]]; then
-    	addon_payload="{\"addon\":{\"id\":\"${addon_id}\"}, \"parameters\": { \"items\": [{\"id\": \"cidr-range\", \"value\": \"10.1.0.0/26\"}, {\"id\": \"addon-managed-api-service\", \"value\": \"${QUOTA}\"}] }}"
+    	addon_payload="{\"addon\":{\"id\":\"${addon_id}\"}, \"parameters\": { \"items\": [{\"id\": \"cidr-range\", \"value\": \"10.1.0.0/26\", \"id\": \"addon-resource-required\", \"value\": true }, {\"id\": \"addon-managed-api-service\", \"value\": \"${QUOTA}\"}] }}"
     fi
 
     echo "Applying ${addon_id} Add-on on a cluster with ID: ${cluster_id}"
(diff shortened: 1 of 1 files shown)


psturc:
👀

scripts/ocm/ocm.sh
```diff
@@ -157,7 +157,7 @@ install_addon() {
 
     # Add mandatory "cidr-range" and "addon-managed-api-service" (quota) params with default value in case of rhoam (managed-api-service) addon 
     if [[ "${addon_id}" == "managed-api-service" ]]; then
-    	addon_payload="{\"addon\":{\"id\":\"${addon_id}\"}, \"parameters\": { \"items\": [{\"id\": \"cidr-range\", \"value\": \"10.1.0.0/26\"}, {\"id\": \"addon-managed-api-service\", \"value\": \"${QUOTA}\"}] }}"
+    	addon_payload="{\"addon\":{\"id\":\"${addon_id}\"}, \"parameters\": { \"items\": [{\"id\": \"cidr-range\", \"value\": \"10.1.0.0/26\", \"id\": \"addon-resource-required\", \"value\": true }, {\"id\": \"addon-managed-api-service\", \"value\": \"${QUOTA}\"}] }}"
```
psturc, line 160: I haven't tested it, but should it look like this
```suggestion
    	addon_payload="{\"addon\":{\"id\":\"${addon_id}\"}, \"parameters\": { \"items\": [{\"id\": \"cidr-range\", \"value\": \"10.1.0.0/26\"}, {\"id\": \"addon-resource-required\", \"value\": true }, {\"id\": \"addon-managed-api-service\", \"value\": \"${QUOTA}\"}] }}"
```

psturc: Actually I just tested it out and it seems that also parameter's value have to be always a type of string
```
{
  "kind": "Error",
  "id": "400",
  "href": "/api/clusters_mgmt/v1/errors/400",
  "code": "CLUSTERS-MGMT-400",
  "reason": "Bad request body: addon parameter(s) 'value' must be of type 'string'",
  "operation_id": "5c937a88-d283-46dd-906b-3d975278a7fa"
}
```
so
```suggestion
    	addon_payload="{\"addon\":{\"id\":\"${addon_id}\"}, \"parameters\": { \"items\": [{\"id\": \"cidr-range\", \"value\": \"10.1.0.0/26\", \"id\": \"addon-resource-required\", \"value\": "true" }, {\"id\": \"addon-managed-api-service\", \"value\": \"${QUOTA}\"}] }}"
```

austincunningham: :+1:

psturc: (just realized I forgot about backslashes 🤦  updated ^)

psturc:
tested via pipeline

/lgtm
/approve

PR #211, opened 2021-06-03
Added CLI command to get supported versions

Boomatang:
# Description
This command downloads managed-tenants to a tmp folder and returns a list of supported version of an operator.
The .gitkeep files are to ensure the empty directories are kept.

Some unit tests require access to gitlab on the vpn. If these test fail to get access then the test is skipped.

**JIRA:** https://issues.redhat.com/browse/MGDAPI-1736

# Validation 
- build this branch locally `make build/cli`
  - Go version is 1.13.x
- run `./delorean pipeline supported-versions`
  - Expected: List of supported versions for rhoam
- run `./delorean pipeline supported-versions --olm-type=integreatly-operator`
  - Expected: List of supported versions for rhmi
- run `./delorean pipeline supported-versions --help`
  - Explore the help for the command and the other flags that can be passed in. 
  - Check that the defaults are as expected.

### DIFF
--- a/cmd/getSupportedVersions.go
+++ b/cmd/getSupportedVersions.go
@@ -1,0 +1,262 @@
+package cmd
+
+import (
+	"context"
+	"fmt"
+	"github.com/blang/semver"
+	"github.com/go-git/go-git/v5"
+	"github.com/integr8ly/delorean/pkg/types"
+	"github.com/spf13/cobra"
+	"io/ioutil"
+	"os"
+	"path"
+	"path/filepath"
+	"sort"
+	"strings"
+)
+
+type getSupportedVersionsFlags struct {
+	olmType                string
+	supportedMajorVersions string
+	supportedMinorVersions string
+	managedTenants         string
+}
+
+type getSupportedVersionsCmd struct {
+	olmType                string
+	supportedMajorVersions int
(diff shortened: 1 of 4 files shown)


cmd/getSupportedVersions.go
```diff
@@ -0,0 +1,262 @@
...
+}
+
+func (c *getSupportedVersionsCmd) run(ctx context.Context) ([]string, error) {
+	olmTypePath, err := getOlmTypePath(c.olmType)
+	if err != nil {
+		return nil, err
+	}
+
+	// Download the manged tenants repo
+	repoDir, err := downloadManagedTenants(c.manageTenants)
```
MStokluska, line 84: WDYT of using this instead -> https://github.com/integr8ly/delorean/blob/master/cmd/createProdsecManifest.go#L137 - it has been unit tested and is widely used across other commands.

Boomatang: I have being looking at that again. The is one issue with the method you linked too. There is no way to turn off the outputing the progress to standard out. This is something we where wanting to do.

MStokluska: Good point, I agree.

MStokluska:
/lgtm
Looks great! Good job!

PR #212, opened 2021-06-04
update to latest prow config

laurafitzgerald:
Updating the version of the file which is used for generating prow config for new releases.
I've updated the Red Hat Managed Integration one also even though they haven't bumped the go version yet.
We'll likely need to update that one again after the next release if they pull in the go version bump.

cc @MStokluska @pmccarthy

### DIFF
--- a/cmd/openshiftCIAddBranch.go
+++ b/cmd/openshiftCIAddBranch.go
@@ -12,8 +12,8 @@ import (
 )
 
 const (
-	ProwConfigSourceRHMI   = "ci-operator/config/integr8ly/integreatly-operator/integr8ly-integreatly-operator-release-v2.0.yaml"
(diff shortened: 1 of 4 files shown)


MStokluska:
@laurafitzgerald it looks good to merge but I think unit tests need to be updated!

laurafitzgerald:
should be good now @MStokluska 👍

MStokluska:
/lgtm
Code looks fine and the unit tests pass, no ways of verifying it otherwise.

laurafitzgerald:
thanks for the lgtm @MStokluska 
This functionality would have been verified with previous release pipelines. The only thing to verify here is that the files which we want to copy exist in the release repo which can be seen at the links below
https://github.com/openshift/release/blob/master/ci-operator/config/integr8ly/integreatly-operator/integr8ly-integreatly-operator-release-v2.8.yaml
https://github.com/openshift/release/blob/master/ci-operator/config/integr8ly/integreatly-operator/integr8ly-integreatly-operator-rhoam-release-v1.7.yaml
will share in the channel for approval.

PR #215, opened 2021-06-10
OSD upgrade pipeline changes + refactoring

psturc:
### JIRA
https://issues.redhat.com/browse/MGDAPI-196

### Changes
* fixed the OSD upgrade function
* added support for updating (stable) upgrade channel (in case the upgrade requires it)

### To verify (z-stream and y-stream upgrade)
1. Install OSD cluster with version `4.6.27`
2. Get your cluster's credentials:
```bash
OCM_CLUSTER_ID=<your-clusters-id> scripts/ocm/ocm.sh save_cluster_credentials
```
3. Trigger the z-stream upgrade:
```bash
TO_VERSION=latest ./scripts/ocm/ocm.sh upgrade_cluster
```
4. Verify it finished successfully (it could take around 1 hour to finish)
5. Trigger y-stream upgrade:
```bash
TO_VERSION=4.7.19 ./scripts/ocm/ocm.sh upgrade_cluster
```
6. Verify it finished successfully

### DIFF
--- a/scripts/ocm/ocm.sh
+++ b/scripts/ocm/ocm.sh
@@ -156,7 +156,6 @@ install_addon() {
     completion_phase="${2}"
 
     : "${USE_CLUSTER_STORAGE:=true}"
-    : "${PATCH_CR_AWS_CM:=true}"
     : "${WAIT:=true}"
     : "${QUOTA:=20}"
     cluster_id=$(get_cluster_id)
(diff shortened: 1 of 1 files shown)


scripts/ocm/ocm.sh
```diff
@@ -250,19 +243,49 @@ delete_cluster() {
...
+    fi
+
+    upgradesAvailable=$(ocm get cluster "${cluster_id}" | jq -r '.version.available_upgrades | values')
     
-    if [[ $upgradeAvailable == true ]]; then
-        oc --kubeconfig "${CLUSTER_KUBECONFIG_FILE}" adm upgrade --to-latest=true
+    if [[ $upgradesAvailable != "" ]]; then
+        channel_version=$(get_channel_version)
+        echo "Current version of cluster's upgrade channel is ${channel_version}"
+        if [[ $TO_VERSION == "latest" || $TO_VERSION == "${channel_version}."* ]]; then
```
trepel, line 263: I don't think I understand this condition. The TO_VERSION is either "latest" or  something like "4.7.10". The `channel_version` is something like "stabel-4.7". It seems that by "latest" it is actually meant "latest in current channel" - is it intentional or not?

However, if I am not mistaken the 2nd part of the condition `$TO_VERSION == "${channel_version}."*` always evaluates to 'false' since string like "4.7.10" never matches string like "stable-4.7". Or am I wrong?

trepel: Ah, my bad, I totally missed the `printf "%s" "${channel_spec##*-}"` line, so `channel_version` is not like "stable-4.7", but it is like "4.7". So the expression does make sense to me now.

PR #217, opened 2021-06-18
MGDAPI-1096 Update ocm.sh script with parameter for cluster node type

jjkiely:
Closes MGDAPI-1096
Allows pipelines to specify node type when provisioning a cluster with ocm.sh script

### DIFF
--- a/scripts/ocm/ocm.sh
+++ b/scripts/ocm/ocm.sh
@@ -63,6 +63,7 @@ create_cluster_configuration_file() {
     : "${PRIVATE:=false}"
     : "${MULTI_AZ:=false}"
     : "${COMPUTE_NODES_COUNT:=}"
+    : "${COMPUTE_MACHINE_TYPE:=}"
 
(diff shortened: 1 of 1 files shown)


trepel:
Tested via:
```
make ocm/cluster.json COMPUTE_MACHINE_TYPE=m5.2xlarge COMPUTE_NODES_COUNT=3 MULTI_AZ=false > cluster.json
make ocm/cluster/create
```

The cluster with 3 worker nodes of m5.2xlarge ec2 type was indeed created.

I just think that the description of `COMPUTE_NODES_COUNT`  is wrong. It says `(default: Configured based on quota for rhoam, can be specified otherwise)`, but I think the reality is that the default value is 8 (see https://github.com/integr8ly/delorean/blob/b655e1732ac32241cca2561f0305ebee9c5cad99/templates/ocm/cluster-template.json#L16

Adding lgtm anyway, but I would defer to @psturc to approve.

/lgtm

jjkiely:
Thanks for the feedback @trepel I've addressed your comment could you please re-add the lgtm label when you get the chance :slightly_smiling_face:

PR #221, opened 2021-07-09
feat: productTests: namespace deletion is optional

psturc:
https://issues.redhat.com/browse/MGDAPI-2176

### How to test

```bash
# build delorean binary
make build/cli

# create test-config file with only A01 test to be run
cat > test-config.yaml <<EOF
---

tests:
- name: integreatly-operator-test
  image: quay.io/integreatly/integreatly-operator-test-harness:master
  regExpFilter: "A01"
  timeout: 7200
  envVars:
  - name: DESTRUCTIVE
    value: 'false'
  - name: MULTIAZ
    value: 'false'
  - name: WATCH_NAMESPACE
    value: redhat-rhoam-operator

# After running this, test-functional namespace should be removed after test container is finished
./delorean pipeline product-tests --test-config test-config.yaml --output test-results --namespace test-functional --post-cleanup

# When this command is finished, test-functional namespace should remain on the cluster
./delorean pipeline product-tests --test-config test-config.yaml --output test-results --namespace test-functional
```

### DIFF
--- a/cmd/productTests.go
+++ b/cmd/productTests.go
@@ -50,6 +50,7 @@ type runTestsCmdFlags struct {
 	outputDir       string
 	timout          int64
 	namespace       string
+	cleanup         bool
 	testsConfigFile string
 }
 
@@ -58,6 +59,7 @@ type runTestsCmd struct {
 	tests     []*TestContainer
 	outputDir string
 	namespace string
+	cleanup   bool
 	oc        utils.OCInterface
 }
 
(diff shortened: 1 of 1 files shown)


cmd/productTests.go
```diff
@@ -83,6 +85,7 @@ func init() {
 	cmd.Flags().StringVarP(&f.outputDir, "output", "o", "", "Absolute path of the output directory to save reports")
 	cmd.MarkFlagRequired("output")
 	cmd.Flags().StringVarP(&f.namespace, "namespace", "n", defaultNamespace, "The namespace to run the test containers")
+	cmd.Flags().BoolVar(&f.cleanup, "post-cleanup", false, "Delete the namespace after test containers finish")
```
psturc, line 88: I set this to `false`, because I don't see a reason why we can't leave the testing namespace present after tests are done. 

If we would want to reuse the namespace again (for another round of test run, for example in upgrade pipeline, where we run the tests pre and post upgrade), we could update our pipelines to always delete this namespace before running the tests.

psturc:
@Boomatang let me know what you think

Boomatang:
@psturc This looks good to me. I have ran the steps given above and it works as expected.

Also tried running the commands give when the test-functional was already existing. In this case the container would fail to start. This I don't think is an issue for this PR but do we want to put in some kind checks to see if the test namespace is already existing?

PR #222, opened 2021-07-27
updating yq to v4

Patryk-Stefanski:
# Description

[MGDAPI-1147 ](https://issues.redhat.com/browse/MGDAPI-1147l)  Update of yq to v4 in integreatly operator meant all related repositories and files also have to be updated to yq v4+.

### Changes Made 

- Yq updated from version 3 to version 4.9.8(most recent at the time) .
- scripts in the following files have been removed as there was no reference to the scripts in the ci-cd repository : 
   - `scripts/delorean/csv_helper`
   - `scripts/delorean/process-csv-images`
   - `scripts/delorean/process-image-manifests`

### Verification 
  Run the following make command and ensure it matches the expected result : 

- `make image/test`
   Expected result : SUCCESS!


austincunningham (commented):
@Patryk-Stefanski not sure we should be removing these scripts, delorean is not just used for ci-cd there is also cli tooling to be considered.  @psturc do you know the usage for the scripts below?

psturc:
> @Patryk-Stefanski not sure we should be removing these scripts, delorean is not just used for ci-cd there is also cli tooling to be considered. @psturc do you know the usage for the scripts ?

I have no idea. Maybe @mikenairn might remember? ^

laurafitzgerald:
The reason the files were removed was under my suggestion. As far as I recall these were the original versions of the automation but were re written to use the delorean cli commands which are implemented using the go cli in this repo. As far as I know we don't use them anywhere. If they are used somewhere they've been marked as deprecated and the user can reach out to us. we'll reinstate them or advise them to use the delorean cli commands which are more actively maintained.

makslion:
In this case, this change looks good to me. 

/lgtm

scripts/delorean/csv_helper
```diff
@@ -1,18 +1,3 @@
...
-}
-
-get_current_csv_file() {
-    echo "$(grep -lR "name: $(get_current_csv $1)" ${MANIFESTS_DIR}/integreatly-$1)"
-}
-
-get_current_version() {
-    echo "$(yq r "$(get_current_csv_file $1)" spec.version)"
-}
+echo  "This script file has deprecated ,please contact  the RHOAM team for help"
```
laurafitzgerald, line 3:
```suggestion
echo  "This script file has deprecated ,please contact  the RHOAM team for help"
exit 1
```

scripts/delorean/process-csv-images
```diff
@@ -1,128 +1,3 @@
...
-if [ -z "${MANIFESTS_DIR}" ]; then
-    echo "MANIFEST_DIR is not set!!"
-    exit 1
-else
-    process_csv
-    if [[ -s ${MANIFESTS_DIR}/integreatly-${PRODUCT}/image_mirror_mapping ]]; then
-      update_image_pull_secret
-    fi
-fi
+echo  "This script file has deprecated ,please contact  the RHOAM team for help"
```
laurafitzgerald, line 3:
```suggestion
echo  "This script file has deprecated ,please contact  the RHOAM team for help"
exit 1
```

scripts/delorean/process-image-manifests
```diff
@@ -1,209 +1,3 @@
...
-   update_package_yaml
-}
-
-if [ -z "${MANIFESTS_DIR}" ]; then
-    echo "MANIFEST_DIR is not set!!"
-    exit 1
-else
-    process_image
-fi
+echo  "This script file has deprecated ,please contact  the RHOAM team for help"
```
laurafitzgerald, line 3:
```suggestion
echo  "This script file has deprecated ,please contact  the RHOAM team for help"
exit 1
```

laurafitzgerald (commented):
@Patryk-Stefanski I was thinking about this one a little more. Given we are exiting with an error which should tell the user that the scripts are no longer available we should error out so _if_ there is someone trying to use it in a pipeline that it will fail in order to indicate there is some problem.

laurafitzgerald:
Thanks for addressing the comments. I wasn't able to run the full verification steps but I believe that was due to my local setup. The image itself was built and yq version is correctly outputted.
/lgtm
/approve

PR #223, opened 2021-08-09
fix: polarionImport: ignore archives without junit

psturc:
### What

If Jenkins pipeline doesn't manage to create JUnit file (can be caused by a failure when fetching test results, issue with Jenkins slave etc.), it will be now ignored and the archive with test results is marked as processed.


trepel:
I don't understand the polarion import much but afaict marking the archive that does not contain any results as processed makes sense to me.

/lgtm

PR #224, opened 2021-08-26
feat: pipeline cleanup-aws

psturc:
### Description
This command takes care of removing unused resources from AWS

* remove "managed-velero"-prefixed s3 buckets, that are no longer used by any OSD cluster
* remove empty (untagged) VPCs (these sometimes get created by cloud-resource-operator) and cluster VPCs that are no longer associated with EC2 instance
* remove all RHMI/RHOAM-tagged resources that are no longer connected to any OSD cluster. Use cluster-service to do that (cluster-service also deletes RDS, Elasticache and all related resources)

### How to use
1. Clone this branch
2. Run 
```bash
make build/cli
# Use AWS account used for creating OSD clusters and installing RHMI/RHOAM
export AWS_ACCESS_KEY_ID=<replace_me> AWS_SECRET_ACCESS_KEY="<replace_me>" 
export AWS_REGION=<replace_me>
# Run the command in dry-run mode, which will only list the resources that are about to be deleted
./delorean pipeline cleanup-aws --region $AWS_REGION --dry-run=true
# Run the same command in debug mode, to see more logs
./delorean pipeline cleanup-aws --region $AWS_REGION --dry-run=true --debug
# Run the command without dry-run to delete the resources
./delorean pipeline cleanup-aws --region $AWS_REGION --dry-run=false
# Run the command without dry-run again to verify that previously lister resources were deleted
./delorean pipeline cleanup-aws --region $AWS_REGION --dry-run=false
```

### To do
- [x] unit tests


Boomatang (commented):
Code looks good, just a few very small points. Going to do the verification steps now

cmd/cleanupAwsAccount.go
```diff
@@ -0,0 +1,380 @@
...
+
+		}
+	}
+
+	log.Info("Following OSD cluster resources were not deleted")
+	for _, awsResources := range c.osdResources {
+		for _, j := range awsResources {
+			log.Infof("Resource type: %s, ID: %s, Tags: %+v\n", j.resourceType, j.ID, j.tags)
+		}
+	}
```
Boomatang, line 193: After doing the verification steps it seemed strange to get this log message when there was no resources that were not deleted. Do you want to put this into a check to see if there is any resources left before logging the information.

psturc: Yes that makes sense. This is now addressed in a new commit.

cmd/cleanupAwsAccount.go
```diff
@@ -0,0 +1,380 @@
...
+func (c *cleanupAwsAccountCmd) fetchS3Buckets(ctx context.Context) error {
+	log.Debug("Fetch S3 Buckets")
+
+	fetchedBuckets, err := c.s3.ListBucketsWithContext(ctx, &s3.ListBucketsInput{})
+	if err != nil {
+		return fmt.Errorf("failed to list s3 buckets: %v", err)
+	}
+
+	for _, b := range fetchedBuckets.Buckets {
+		bucketLocation, err := c.s3.GetBucketLocation(&s3.GetBucketLocationInput{Bucket: b.Name})
```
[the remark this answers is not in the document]
psturc: fixed

cmd/cleanupAwsAccount.go
```diff
@@ -282,13 +288,13 @@ func (c *cleanupAwsAccountCmd) fetchVpcs(ctx context.Context) error {
 		for _, tag := range vpc.Tags {
 			newVpc.tags[aws.StringValue(tag.Key)] = aws.StringValue(tag.Value)
 		}
-		newVpc.clusterTag = extractClusterTag(newVpc.tags)
-
-		if newVpc.clusterTag != "" {
-			c.rhmiResources[newVpc.clusterTag] = append(c.osdResources[newVpc.clusterTag], newVpc)
+		var hasRhmiTag bool
+		newVpc.clusterTag, hasRhmiTag = extractClusterTag(newVpc.tags)
```
Boomatang, line 292: Would this be cleaner here and remove the `var hasRhmiTag bool`
```
newVpc.clusterTag, hasRhmiTag := extractClusterTag(newVpc.tags)
```

Boomatang (commented):
Small code comments. Going to rerun the verification steps and run the new unit test

cmd/cleanupAwsAccount.go
psturc: That was my initial plan. However that does not work for a reason I do not understand 🤷‍♂️ 
Screen Shot 2021-09-10 at 12 38 15

cmd/cleanupAwsAccount.go
```diff
@@ -419,14 +427,17 @@ func isVeleroBucket(bucketName string) bool {
...
+			return value, true
 		}
-		if strings.Contains(key, rhmiResourceTagPrefix) || strings.Contains(key, managedVeleroTagPrefix) {
-			return value
+		if strings.Contains(key, managedVeleroTagPrefix) {
+			return value, false
 		}
 	}
-	return ""
+	return
```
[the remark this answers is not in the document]
psturc: Sorry if it's confusing, but [in golang it's actually not needed to specify which values you return in case you name them](https://tour.golang.org/basics/7)

I wanted to give these return values some names, so it's more clear what the values (especially the second one) mean.

cmd/cleanupAwsAccount.go
psturc: found [an explanation for this](https://stackoverflow.com/a/52085092/3894898)

cmd/cleanupAwsAccount.go
```diff
@@ -260,81 +282,90 @@ func (c *cleanupAwsAccountCmd) fetchVpcs(ctx context.Context) error {
...
 		if isVeleroBucket(b.ID) {
-			if !c.containsActiveOsdClusterTag(b.tags) {
+			if _, ok := c.osdResources[b.clusterTag]; !ok {
 				if c.dryRun {
-					log.Info("would delete a bucket: ", b.ID)
+					log.Infof("would delete a bucket '%s' (tag: '%s')\n", b.ID, b.clusterTag)
 				} else {
 					if err := c.removeS3Bucket(b.ID, ctx); err != nil {
-						return err
+						log.Warnf("Failed to delete S3 bucket '%s' (tag '%s'). It might be already deleted\n", b.ID, b.clusterTag)
```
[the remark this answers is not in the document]
psturc: I've added a `log.Debug` for every error that is not critical for the cleanup command, i.e. clean up command can still finish successfully, if such error appears.

Boomatang:
I have ran verification steps, Everything seems to be in good order. Unit test are passing. Any questions I might have had are after being addressed. 
/lgtm

PR #225, opened 2021-08-31
OCM script - check for existing cluster

echernous:
**Description**
Add a check if there a cluster already exists with the specified cluster name, if it exists get existing cluster details and proceed further, otherwise, send a request to create a new cluster. I wasn't sure if it's a breaking change, pls could you advise and I can add an extra variable which can indicate if the existing cluster should be used or ignored as it was before. Thank you!

**Use Case**
As a User I would like to use this script in a Jenkins parallel job so when several jobs execute the create_cluster method with the same cluster name they would use the same cluster. In this case I don't need to wait for the first job to finish before triggering other jobs. 

**Testing**
1. Tried to execute new cluster creation using unique cluster name - new cluster was created.
2. Tried to create a new cluster using the same name from step 1 - it picked up existing cluster details and proceeded to wait for its ready status.

### DIFF
--- a/scripts/ocm/ocm.sh
+++ b/scripts/ocm/ocm.sh
@@ -83,7 +83,7 @@ create_cluster_configuration_file() {
     jq ".expiration_timestamp = \"${timestamp}\" | .name = \"${OCM_CLUSTER_NAME}\" | .display_name = \"${cluster_display_name}\" | .region.id = \"${OCM_CLUSTER_REGION}\" | .api.listening = \"${listening}\"" \
         < "${CLUSTER_TEMPLATE_FILE}" \
         > "${CLUSTER_CONFIGURATION_FILE}"
-	
+
     if [ "${BYOC}" = true ]; then
         check_aws_credentials_exported
         update_configuration "aws"
@@ -105,7 +105,7 @@ create_cluster_configuration_file() {
         update_configuration "compute_machine_type"
     fi
 
-    
+
 
 
     cat "${CLUSTER_CONFIGURATION_FILE}"
@@ -161,7 +161,7 @@ install_addon() {
     cluster_id=$(get_cluster_id)
     addon_payload="{\"addon\":{\"id\":\"${addon_id}\"}}"
 
-    # Add mandatory "cidr-range" and "addon-managed-api-service" (quota) params with default value in case of rhoam (managed-api-service) addon 
+    # Add mandatory "cidr-range" and "addon-managed-api-service" (quota) params with default value in case of rhoam (managed-api-service) addon
     if [[ "${addon_id}" == "managed-api-service" ]]; then
     	addon_payload="{\"addon\":{\"id\":\"${addon_id}\"}, \"parameters\": { \"items\": [{\"id\": \"cidr-range\", \"value\": \"10.1.0.0/26\"}, {\"id\": \"addon-resource-required\", \"value\": \"true\" }, {\"id\": \"addon-managed-api-service\", \"value\": \"${QUOTA}\"}] }}"
     fi
@@ -256,7 +256,7 @@ upgrade_cluster() {
     fi
 
     upgradesAvailable=$(ocm get cluster "${cluster_id}" | jq -r '.version.available_upgrades | values')
-    
+
     if [[ $upgradesAvailable != "" ]]; then
         channel_version=$(get_channel_version)
         echo "Current version of cluster's upgrade channel is ${channel_version}"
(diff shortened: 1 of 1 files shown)


scripts/ocm/ocm.sh
```diff
@@ -320,7 +320,14 @@ get_infra_id() {
...
     local cluster_details
-    cluster_details=$(ocm post /api/clusters_mgmt/v1/clusters --body="${CLUSTER_CONFIGURATION_FILE}" | jq -r | tee "${CLUSTER_DETAILS_FILE}")
+    ocm_command="ocm post /api/clusters_mgmt/v1/clusters --body='${CLUSTER_CONFIGURATION_FILE}'"
+    # Get existing cluster details if exists to avoid DuplicateClusterName error
+    existing_cluster_id=$(ocm get /api/clusters_mgmt/v1/clusters | jq -r ".items[] | select(.name==\"${OCM_CLUSTER_NAME}\") | .id")
+    if [[ ! -z "${existing_cluster_id:-}" ]]; then
+        ocm_command="ocm get /api/clusters_mgmt/v1/clusters/${existing_cluster_id}"
+    fi
+
+    cluster_details=$(eval ${ocm_command} | jq -r | tee "${CLUSTER_DETAILS_FILE}")
```
agullon, line 330: I'm not sure I understand the flow correctly. What will happen when already exists a cluster with the same as the requested one? I'd expect cluster's details are printed out and execution stops.

echernous: the execution will continue, it will get cluster details from the existing cluster, the parent method that calls this method will print cluster details such as login and continue until the cluster is ready

agullon: understood, I agree with this change.

scripts/ocm/ocm.sh
```diff
@@ -320,7 +320,14 @@ get_infra_id() {
 
 send_cluster_create_request() {
     local cluster_details
-    cluster_details=$(ocm post /api/clusters_mgmt/v1/clusters --body="${CLUSTER_CONFIGURATION_FILE}" | jq -r | tee "${CLUSTER_DETAILS_FILE}")
+    ocm_command="ocm post /api/clusters_mgmt/v1/clusters --body='${CLUSTER_CONFIGURATION_FILE}'"
+    # Get existing cluster details if exists to avoid DuplicateClusterName error
+    existing_cluster_id=$(ocm get /api/clusters_mgmt/v1/clusters | jq -r ".items[] | select(.name==\"${OCM_CLUSTER_NAME}\") | .id")
```
psturc, line 325: got an error when running the cluster/create command:

```bash
➜  delorean git:(ocm-use-existing-cluster) make ocm/cluster.json OCM_CLUSTER_LIFESPAN=36 BYOC=true OCM_CLUSTER_NAME=test-psturc OCM_CLUSTER_REGION=us-east-2 MULTI_AZ=false COMPUTE_MACHINE_TYPE=m5.2xlarge COMPUTE_NODES_COUNT=3
...
➜  delorean git:(ocm-use-existing-cluster) make ocm/cluster/create
scripts/ocm/ocm.sh: line 325: OCM_CLUSTER_NAME: unbound variable
Error: Can't print body: signal: broken pipe
make: *** [ocm/cluster/create] Error 1
```

psturc: I would prefer to read the cluster name value from the `cluster.json` file instead of the variable, wdyt?

echernous: @psturc Yes, I agree, in case users don't run the create configuration command first the create cluster command will fail. Let me check and I will come back to you. Thanks!

echernous: Hey @psturc added `get_cluster_name`, pls could you check when have a chance? Thank you!

scripts/ocm/ocm.sh
```diff
@@ -320,7 +324,17 @@ get_infra_id() {
 
 send_cluster_create_request() {
     local cluster_details
-    cluster_details=$(ocm post /api/clusters_mgmt/v1/clusters --body="${CLUSTER_CONFIGURATION_FILE}" | jq -r | tee "${CLUSTER_DETAILS_FILE}")
+    local existing_cluster_id
+    local ocm_command
+
+    ocm_command="ocm post /api/clusters_mgmt/v1/clusters --body='${CLUSTER_CONFIGURATION_FILE}'"
+    # Get existing cluster details if exists to avoid DuplicateClusterName error
+    existing_cluster_id=$(ocm get /api/clusters_mgmt/v1/clusters | jq -r ".items[] | select(.name==\"$(get_cluster_name)\") | .id")
```
psturc, line 332: @echernous sorry, I tried to run `make ocm/cluster/create` with a config file for a cluster that already exists, but it doesn't work due to this command (it does not return anything):
```bash
➜  delorean git:(ocm-use-existing-cluster) ocm get /api/clusters_mgmt/v1/clusters | jq -r '.items[] | select(.name=="test-rhoam-1191") | .id'
```

I usually use following command to query ocm API. it works every time (we use it also in our pipelines to fetch cluster details):
```bash
➜  delorean git:(ocm-use-existing-cluster) ocm get clusters --parameter search="name like 'test-rhoam-1191'" | jq -r '.items[0].id'
1mtr0h8ger42afe6hdju6jggtofgt6en
```

Could you please try if it works on your machine as well and if so, replace this line with it? Thanks

echernous: @psturc strange, it was working for me locally and on our jenkins server. I changed the command to get existing cluster ids and tried locally and triggering jenkins jobs, seems working.

psturc:
```bash
➜  delorean git:(ocm-use-existing-cluster) make ocm/cluster/create
Info: Cluster with the given name already exists, continue with the existing cluster details
Cluster ID: 1mtr0h8ger42afe6hdju6jggtofgt6en
Waiting for cluster creation for 180m...
```

Verified. Thanks @echernous for this improvement!

There's just one PR check that's failing and need to be addressed before merging.

You can run `make test/lint` to see the details.

echernous:
@psturc fixed, thanks for your help and prompt review

PR #230, opened 2021-09-15
added a way to install RHOAM on OSD trial cluster using addon-flow

martinhesko:
had to add a new function to differentiate between an install on regular and trial cluster, for some reason they use different parameter names for their quota.

Tested out successfully on my local jenkinks


trepel:
Looks good to me. I am going to verify via pipeline just to be sure.

trepel:
Tested via Jenkins, looks good.

/lgtm

psturc:
code /lgtm
also confirmed with @martinhesko  that the `install_rhoam` function still works properly.

/approve

PR #231, opened 2021-09-22
fix: ocm: typo

psturc:
Hi @echernous, I just noticed an error when I tried to create a cluster with cluster id that was already there. Not sure how I missed it in the original PR verification :)

I assume that "[`// empty`](https://github.com/integr8ly/delorean/compare/master...psturc:ocm-fix-typo?expand=1#diff-b26057c72b176f72b860716e58baec8c2a636cfd8c12b40bc199ff6c16653892L329)" was a typo/put there by accident?

### DIFF
--- a/scripts/ocm/ocm.sh
+++ b/scripts/ocm/ocm.sh
@@ -326,7 +326,7 @@ get_cluster_region() {
 }
 
 get_existing_cluster_id() {
-    ocm get clusters --parameter search="name like '$(get_cluster_name)'" | jq -r '.items[0].id // empty'
+    ocm get clusters --parameter search="name like '$(get_cluster_name)'" | jq -r '.items[0].id | values'
 }
 
 is_ccs_cluster() {


echernous:
Hi @psturc I remember adding it on purpose so the command would return an empty line instead of null, otherwise, it was failing in bash if I run the script. Let me test it today and come back with results. The current script version works without issues, we already integrated in into our pipeline.

psturc:
@echernous hmm you're right, when I tried it again with another cluster (with code from master branch), it worked this time...

```bash
➜  delorean git:(master) make ocm/cluster/create
2021/09/22 14:58:27 protocol error: received *http.http2GoAwayFrame before a SETTINGS frame
{
  "kind": "Error",
  "id": "400",
  "href": "/api/clusters_mgmt/v1/errors/400",
  "code": "CLUSTERS-MGMT-400",
  "reason": "Duplicate cluster name. There is already a cluster with the name 'xxx' in the organization 'xxx'",
  "details": [
    {
      "Error_Key": "DuplicateClusterName"
    }
  ],
  "operation_id": "xxx"
}
```

So the error has probably something to do with this: `2021/09/22 14:58:27 protocol error: received *http.http2GoAwayFrame before a SETTINGS frame`, but not sure what caused it.

I will leave this PR open and close it in future if it's not needed.

PR #232, opened 2021-10-06
MGDAPI-2561: Adapt CSV processing for multiple containers

sergioifg94:
The `ews process-csv` command processes the CSV to generate an image mirror mapping file. In order to do this it finds image references in relevant fields, including the containers field of the operator Deployment. In the past, operators would have only one container (the controller manager itself), but newer operators, since the release of operator-sdk 1.0, include a proxy container.

The CSV processing logic would take only the first container for
processing, thi change iterates through the containers list and
processes them all.

Adapt the unit tests to consider this situation

### DIFF
--- a/pkg/utils/csv.go
+++ b/pkg/utils/csv.go
@@ -519,17 +519,22 @@ func GetAndUpdateOperatorImage(manifestDir string, images map[string]string, isG
 			annotations["containerImage"] = prodImage
 		}
 	} else {
-		//check for stage and prod images and replace with quay image
-		stageMatched := ReStage.FindString(deployment.Spec.Template.Spec.Containers[0].Image)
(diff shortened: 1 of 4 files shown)


pkg/utils/csv.go
```diff
@@ -539,7 +544,7 @@ func GetAndUpdateOperatorImage(manifestDir string, images map[string]string, isG
 	csv.SetOperatorDeploymentSpec(deployment)
 
 	err = csv.WriteYAML(fp)
-	return images, nil
+	return images, err
```
MStokluska, line 547: hey Sergio, I might be missing something but I'm trying to understand this change, if we get here there would be no err right? as other 2 error checks are returning nil, err. Just out of curiosity, if we know that at this stage error can't be anything else than nil, is it still better to return err? I presume this is just for visibility?

PR #237, opened 2021-11-10
Added a way to pass arguments to the remote container

martinhesko:
This PR is related to the main PR that can be found[here](https://gitlab.cee.redhat.com/integreatly-qe/ci-cd/-/merge_requests/593)

This way all we need to run a different set of tests in the 3scale test suite is to change a line in the .yaml configuration file

### DIFF
--- a/cmd/productTests.go
+++ b/cmd/productTests.go
@@ -37,7 +37,7 @@ type TestContainer struct {
 	ImagePullSecret string      `json:"ImagePullSecretEnvVar,omitempty"`
 	EnvVars         []v1.EnvVar `json:"envVars,omitempty"`
 	RegExpFilter    string      `json:"regExpFilter,omitempty`
-    Argument        string      `json:"argument, omitempty`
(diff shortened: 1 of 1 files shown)


cmd/productTests.go
```diff
@@ -326,3 +327,9 @@ func (c *runTestsCmd) completeJob(pod v1.Pod) error {
 func parseSecretName(pullSecret string) string {
 	return strings.ToLower(strings.ReplaceAll(pullSecret, "_", "-"))
 }
+func assignArguments(t *TestContainer) []string {
+	if t.Argument != "" {
+		return []string{t.Argument}
+	}
+	return []string{"-ginkgo.focus", t.RegExpFilter}
```
psturc, line 334: thanks, I created a [follow-up ticket](https://issues.redhat.com/browse/MGDAPI-3086) to use the "argument" param instead of "regExpFilter" by our pipelines for running functional tests too

PR #239, opened 2021-11-29
fix: aws cleanup: ignore terminated ec2 instances

psturc:
### Description

If the cleanup command runs while related OSD cluster's EC2 instances are still present in AWS (but in `terminated` state), they are still counted as resources that should block the deletion of RHOAM AWS resources (which is incorrect).

These terminated EC2 instances are removed by AWS after a while, so the cleanup command would do the work correctly after they're gone (usually after couple of hours).

This PR contains a fix which allows deleting RHOAM AWS resources immediately right after all OSD cluster EC2 instances are terminated.

### Verification

Verified against the cluster that was freshly terminated while related RHOAM resources were still present in AWS


psturc:
Hi @Boomatang this is a small improvement for the AWS cleanup command. Please take a look and add lgtm if it looks ok, thanks!

Boomatang:
This seems to be working as expected

/lgtm

PR #250, opened 2022-02-11
MGDAPI-3430 

den-rgb:
# Issue link
https://issues.redhat.com/browse/MGDAPI-3185 

# What
Changed default node amount to 5 from 8. Added a user prompt to be able to change the default value if the cluster is being created locally.

# Verification steps
- Run
```
ocm login --url=stage --token=<token>
```
Token found [here](https://console.redhat.com/openshift/token)
```
make ocm/cluster.json
```
```
 make ocm/cluster/create
```
- Verify a prompt appears.
- Verify cluster was created with the amount of nodes you specified.

### DIFF
--- a/scripts/ocm/ocm.sh
+++ b/scripts/ocm/ocm.sh
@@ -27,6 +27,8 @@ readonly ERROR_MISSING_CLUSTER_ID="ERROR: OCM_CLUSTER_ID was not specified"
 
 readonly WARNING_CLUSTER_HEALTH_CHECK_FAILED="WARNING: Cluster was not reported as healthy. You might expect some issues while working with the cluster."
 
+LOCAL_RUN=${LOCAL_RUN:-true}
+
 check_aws_credentials_exported() {
     if [[ -z "${AWS_ACCOUNT_ID:-}" || -z "${AWS_SECRET_ACCESS_KEY:-}" || -z "${AWS_ACCESS_KEY_ID:-}" ]]; then
         printf "%s\n" "${ERROR_MISSING_AWS_ENV_VARS}"
@@ -324,6 +326,10 @@ get_cluster_name() {
     jq -r .name < "${CLUSTER_CONFIGURATION_FILE}"
 }
 
+get_cluster_node_count() {
+    jq -r .nodes.compute < "${CLUSTER_TEMPLATE_FILE}"
+}
+
 get_cluster_subscription_id() {
     jq -r .subscription.id < "${CLUSTER_DETAILS_FILE}"
 }
@@ -354,6 +360,26 @@ send_cluster_create_request() {
     local existing_cluster_id
     local ocm_command
 
+    LOOP=${LOOP:-true}
+    tmp=$(mktemp)
+
+    if [ "$LOCAL_RUN" = true ]; then 
+       while [ "$LOOP" = true ]; 
+       do 
+       read -p "Will a cluster with $(get_cluster_node_count) nodes be enough (Y/N)? " -n 1 -r 
+       if [[ $REPLY =~ ^[Yy]$ ]]; then 
+          echo $'\nCluster with '"$(get_cluster_node_count)"' nodes will be created.' 
+          LOOP=false 
+       elif [[ "$REPLY" =~ ^[Nn]$ ]]; then 
+          echo "" 
+          read -p "How many nodes would you like? " -n 1 -r 
+          echo $'\nCluster with '"$REPLY"' nodes will be created.' 
+          jq '.nodes.compute = '"${REPLY}"'' "${CLUSTER_CONFIGURATION_FILE}" > "$tmp" && mv "$tmp" "${CLUSTER_CONFIGURATION_FILE}" 
+          LOOP=false 
+        else echo $'\nPlease input either "Y" or "N"' 
+       fi 
+       done 
+    fi
     ocm_command="ocm post /api/clusters_mgmt/v1/clusters --body='${CLUSTER_CONFIGURATION_FILE}'"
     # Get existing cluster details if exists to avoid DuplicateClusterName error
     existing_cluster_id=$(get_existing_cluster_id)
@@ -486,7 +512,7 @@ Optional exported variables:
 - OPENSHIFT_VERSION                 to get OpenShift versions, run: ocm cluster versions
 - PRIVATE                           Cluster's API and router will be private
 - MULTI_AZ                          true/false (default: false)
-- COMPUTE_NODES_COUNT               number of cluster's compute nodes (default: 8 in cluster-template. Can be specified otherwise)
+- COMPUTE_NODES_COUNT               number of cluster's compute nodes (default: 5 in cluster-template. Can be specified otherwise)
 - COMPUTE_MACHINE_TYPE              node type of cluster's compute nodes (default: m5.xlarge, can be specified otherwise)
 - OSD_TRIAL                         true/false (default: false)
 ==========================================================================================================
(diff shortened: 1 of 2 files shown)


den-rgb:
@psturc

scripts/ocm/ocm.sh
```diff
@@ -354,6 +360,27 @@ send_cluster_create_request() {
...
+       elif [[ "$ANSWER" = "N" ]]; then 
+          read -p -r "How many nodes would you like?" NEW_NODES_AMOUNT
+          echo Cluster with "$NEW_NODES_AMOUNT" nodes will be created.
+          jq '.nodes.compute = '"${NEW_NODES_AMOUNT}"'' "${CLUSTER_CONFIGURATION_FILE}" > "$tmp" && mv "$tmp" "${CLUSTER_CONFIGURATION_FILE}"
+          LOOP=false
+       else echo "Please input either 'y' or 'N'"
+       fi
+       done
+    fi
+
```
makslion, line 383: Not sure was it something with me, but this logic wasn't working. I've managed to make it behave, be pretty and accept booth upper and lower cases with 
```
    if [ "$LOCAL_RUN" = true ]; then
       while [ "$LOOP" = true ]; do
       read -p "Will a cluster with $(get_cluster_node_count) nodes be enough (Y/N)? " -n 1 -r
       if [[ $REPLY =~ ^[Yy]$ ]]; then
          echo $'\nCluster with '"$(get_cluster_node_count)"' nodes will be created.'
          LOOP=false
       elif [[ "$REPLY" =~ ^[Nn]$ ]]; then
          echo ""
          read -p "How many nodes would you like? " -n 1 -r
          echo $'\nCluster with '"$REPLY"' nodes will be created.'
          jq '.nodes.compute = '"${REPLY}"'' "${CLUSTER_CONFIGURATION_FILE}" > "$tmp" && mv "$tmp" "${CLUSTER_CONFIGURATION_FILE}"
          LOOP=false
       else echo "Please input either 'Y' or 'N'"
       fi
       done
    fi
```
wdyt? :thinking:

makslion:
Hm, should the node number [here](https://github.com/integr8ly/delorean/blob/master/scripts/ocm/ocm.sh#L489) also be updated? Since you are changing the default :thinking:

scripts/ocm/ocm.sh
den-rgb: Yeah looks good I don't see why not.

den-rgb: I'll change the number too.

den-rgb: @makslion I added the changes.

scripts/ocm/ocm.sh
```diff
@@ -354,6 +360,27 @@ send_cluster_create_request() {
...
+       if [[ $REPLY =~ ^[Yy]$ ]]; then 
+          echo $'\nCluster with '"$(get_cluster_node_count)"' nodes will be created.' 
+          LOOP=false 
+       elif [[ "$REPLY" =~ ^[Nn]$ ]]; then 
+          echo "" 
+          read -p "How many nodes would you like? " -n 1 -r 
+          echo $'\nCluster with '"$REPLY"' nodes will be created.' 
+          jq '.nodes.compute = '"${REPLY}"'' "${CLUSTER_CONFIGURATION_FILE}" > "$tmp" && mv "$tmp" "${CLUSTER_CONFIGURATION_FILE}" 
+          LOOP=false 
+        else echo $'\nPlease input either "Y" or "N"'
```
makslion, line 380: is this typo? :thinking:

scripts/ocm/ocm.sh
```diff
@@ -354,6 +360,27 @@ send_cluster_create_request() {
     local existing_cluster_id
     local ocm_command
 
+    # NODE_AMOUNT=$(jq ".nodes.compute " ${CLUSTER_TEMPLATE_FILE})
```
makslion, line 363: Oh, and do we need this comment? Sorry, forget to mention it before

den-rgb: oh no we don't sorry I'll take that out.

scripts/ocm/ocm.sh
den-rgb: what typo are you referencing to?

den-rgb: Are you referring to the extra quotation mark? I don'y know when the test was running one of the errors told me to do it like that.

makslion: Oh, nevermind. That's me being blind. :facepalm: I forced myself to stop seeing `echo echo $'...` here

makslion:
The logic works as expected 
Screenshot from 2022-02-14 16-09-29
The cluster was successfully created with the specified node count. Good work

/lgtm

scripts/ocm/ocm.sh
den-rgb: Oh haha no problem. :+1:

scripts/ocm/ocm.sh
```diff
@@ -354,6 +360,26 @@ send_cluster_create_request() {
     local existing_cluster_id
     local ocm_command
 
+    LOOP=${LOOP:-true}
+    tmp=$(mktemp)
+
+    if [ "$LOCAL_RUN" = true ]; then 
+       while [ "$LOOP" = true ]; 
+       do 
+       read -p "Will a cluster with $(get_cluster_node_count) nodes be enough (Y/N)? " -n 1 -r
```
laurafitzgerald, line 369:
```suggestion
       read -p "Are you sure you need $(get_cluster_node_count) node cluster (Y/N)? " -n 1 -r 
```

laurafitzgerald:
Hi @den-rgb small suggestion inline about the text. We want to encourage the use of smaller clusters if they will do. 👍

makslion:
writing text that would be read by humans is a known issue :smile: now for sure: 

/lgtm

scripts/ocm/ocm.sh
```diff
@@ -324,6 +326,10 @@ get_cluster_name() {
     jq -r .name < "${CLUSTER_CONFIGURATION_FILE}"
 }
 
+get_cluster_node_count() {
+    jq -r .nodes.compute < "${CLUSTER_TEMPLATE_FILE}"
```
psturc, line 330: Since user can provide custom amount of nodes via `COMPUTE_NODES_COUNT` env var, the value from `${CLUSTER_TEMPLATE_FILE}` is not alway correct.

Should be replaced with
`jq -r .nodes.compute < "${CLUSTER_CONFIGURATION_FILE}"`

den-rgb: We use the CLUSTER_TEMPLATE_FILE to reset the node count back to 5, if we were to use the CLUSTER_CONFIGURATION_FILE it would get the node count specified on the previous run.

psturc:
```
➜  delorean git:(master) make ocm/cluster.json COMPUTE_NODES_COUNT=6
Cluster configuration:
{
  "display_name": "rhmi-220217-1057",
  "ccs": {
    "enabled": false
  },
  "managed": true,
  "multi_az": false,
  "name": "rhmi-2202171736",
  "network": {
    "machine_cidr": "10.11.128.0/23",
    "service_cidr": "10.11.0.0/18",
    "pod_cidr": "10.11.64.0/18",
    "host_prefix": 23
  },
  "nodes": {
    "compute": 6,
    "compute_machine_type": {
      "id": "m5.xlarge"
    }
  },
  "region": {
    "id": "eu-west-1"
  },
  "expiration_timestamp": "2022-02-17T14:57:05Z",
  "api": {
    "listening": "external"
  }
}%
➜  delorean git:(master) make ocm/cluster/create
Are you sure you need a 5 node cluster (Y/N)? Please consider a smaller cluster if it will satisfy your development needs.
```

den-rgb: Ahh okay I think I forgot to reset the node count after the user hits "y", I can add in a line to reset it to "get_cluster_node_count"

PR #252, opened 2022-02-24
updated the products test module to accept different entrypoints

martinhesko:
I added this as a quick fix to have the `sh` entrypoint provided to me when running the 3scale products tests, this approach can be revisited later in [MGDAPI-3086](https://issues.redhat.com/browse/MGDAPI-3086)

### DIFF
--- a/cmd/productTests.go
+++ b/cmd/productTests.go
@@ -36,8 +36,9 @@ type TestContainer struct {
 	Timeout         int64       `json:"timeout,omitempty"`
 	ImagePullSecret string      `json:"ImagePullSecretEnvVar,omitempty"`
 	EnvVars         []v1.EnvVar `json:"envVars,omitempty"`
-	RegExpFilter    string      `json:"regExpFilter,omitempty`
-	Argument        string      `json:"argument, omitempty`
+	RegExpFilter    string      `json:"regExpFilter,omitempty"`
+	Entrypoint      []string    `json:"entrypoint, omitempty"`
+	Argument        []string    `json:"argument, omitempty"`
 	Success         bool
 }
 
@@ -263,7 +264,7 @@ func getTestContainerJob(namespace string, t *TestContainer) *batchv1.Job {
 								},
 							},
 							Env:     t.EnvVars,
-							Command: assignEntrypoint(t),
+							Command: t.Entrypoint,
 							Args:    assignArguments(t),
 						},
 						{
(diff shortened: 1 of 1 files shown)


cmd/productTests.go
```diff
@@ -327,9 +327,17 @@ func (c *runTestsCmd) completeJob(pod v1.Pod) error {
 func parseSecretName(pullSecret string) string {
 	return strings.ToLower(strings.ReplaceAll(pullSecret, "_", "-"))
 }
+
+func assignEntrypoint(t *TestContainer) []string {
+	if t.Argument != "" {
```
psturc, line 332: Could we define a new field for the entrypoint value to `TestContainer` struct instead? (i.e. `t.Entrypoint`)

cmd/productTests.go
```diff
@@ -327,9 +327,17 @@ func (c *runTestsCmd) completeJob(pod v1.Pod) error {
...
+		return []string{"sh"}
+	}
+	return []string{"/integreatly-operator-test-harness.test"}
```
psturc, line 335: I think the default value should be empty, because this would override the existing entrypoint in a container, if there is one already defined.

For instance our functional [test container has it](https://github.com/integr8ly/integreatly-operator/blob/11765cfeb32463c15e8a345cec3c5dc5c94ac4dd/Dockerfile.functional#L30), so imo this change would affect the way that our tests are logged in nightly pipelines.

cmd/productTests.go
```diff
@@ -327,9 +327,17 @@ func (c *runTestsCmd) completeJob(pod v1.Pod) error {
...
+}
+
 func assignArguments(t *TestContainer) []string {
 	if t.Argument != "" {
-		return []string{t.Argument}
+		return []string{"-c", fmt.Sprintf("oc project redhat-rhoam-3scale; make %s", t.Argument)}
```
psturc, line 340: would it be possible to take this value from the .yaml config instead?

psturc (changes requested):
Sorry I don't agree with merging current changes.

It would be good if we would avoid hardcoding more things into this command.

I'm aware that there's [one more thing that's also hardcoded (and was introduced by me :) ](https://github.com/integr8ly/delorean/pull/252/files#diff-4ef27691fbd74945fdf80d07c7ebf4d7bcd5a44ef972b48f54c3cfdc4a1d6078R342) but I hope we can address that one too [in this ticket](https://issues.redhat.com/browse/MGDAPI-3086)

cmd/productTests.go
martinhesko: I definitely agree, I defined the Entrypoint in TestContainer as an empty array so if you don't provide an entrypoint it takes the default one from the Dockerfile.

martinhesko:
Thanks for the suggestions, I took them into consideration and I believe the changes I've made look much better.
I'll change the pipeline libraries to wrap entrypoints and arguments into arrays as well.

I tested this out locally to confirm it works when given no entrypoint and that it works regularly by running the 3scale install tools .yaml config on it and the regular 3scale test suite and everything worked as expected.

psturc:
I verified this PR by running 3scale tests via a Jenkins pipeline based on [this MR](https://gitlab.cee.redhat.com/integreatly-qe/ci-cd/-/merge_requests/642) that uses Delorean CLI image built from code in this PR.

I also checked that these changes don't affect nightly pipelines by running functional tests, also via Jenkins pipeline.

/lgtm
/approve

PR #257, opened 2022-03-30
[MGDAPI-2656] add projected service account token

cecobask:


KevFan:
Verified projected service account token is used in pod deployments to allow assuming STS role

/lgtm

PR #259, opened 2022-04-04
MGDAPI-3782 - feat: default to and support ROSA STS on stage

KevFan:


cecobask (approved):
Verified the script works as expected.
/lgtm

PR #260, opened 2022-04-12
MGDAPI-3373 - tagging of images handled by oc mirror/robot accounts

martinhesko:
changed `tagRelease.go` to use robot accounts for tagging images in quay.io

Verified by tagging images in my personal [quay.io](https://quay.io/repository/mhesko0/managed-api-service)

before merging we should make sure that the corresponding quay secret in the pipeline contains a robot account with write permissions to repositories we tag every release.


MStokluska:
@martinhesko any chance we could get unit tests fixed please
```
 cmd/tagRelease_test.go:81:14: undefined: commitIDLabelFilter
cmd/tagRelease_test.go:517:62: cannot use c.quayClient (type *quay.Client) as type string in argument to DoTagRelease 
```

MStokluska:
Vendor test also fails, did you run `make vendor/fix` on it? 
Also, if I could recommend something for future reference, when you are making vendor bumps, it's better to do them in a separate commit as it makes it easier for reviewing, does matter too much in this scenario as it's a single file change but a lot of changes in vendor folder!

austincunningham:
might need a `make vendor/fix`

cmd/tagRelease.go
```diff
@@ -177,46 +195,70 @@ func createGitTag(ctx context.Context, client services.GitService, gitRepoInfo *
...
-	tag := tags.Tags[0]
-	commitID, _, err := quayClient.Manifests.ListLabels(ctx, quayRepo, *tag.ManifestDigest, &quay.ListManifestLabelsOptions{Filter: commitIDLabelFilter})
+func createTagForImage(ctx context.Context, quayRepo string, quaySrcTag string, quayDstTag string, quayTokenFilename string, commitSHA string) error {
+
+	split := strings.Split(quayRepo, "/")
+	quayOrg, quayImage := split[0], split[1]
+
+	buf := new(bytes.Buffer)
+	i := info.NewInfoOptions(genericclioptions.IOStreams{Out: buf})
+	i.Images = append(i.Images, fmt.Sprintf("quay.io/%s/%s:%s", quayOrg, quayImage, quaySrcTag))
```
laurafitzgerald, line 234: Seems like we split the `quayRepo` and then append it together again for i.Images. I think we can skip the split?

cmd/tagRelease.go
```diff
@@ -177,46 +195,70 @@ func createGitTag(ctx context.Context, client services.GitService, gitRepoInfo *
...
+	err := i.Run()
 	if err != nil {
 		return err
 	}
-	if len(commitID.Labels) == 0 {
-		return fmt.Errorf("label %s doesn't exist", commitIDLabelFilter)
-	}
-	if *commitID.Labels[0].Value != commitSHA {
+	var imageInfo ImageInfo
+	json.Unmarshal(buf.Bytes(), &imageInfo)
```
laurafitzgerald, line 243: There's an unhandle error here.

cmd/tagRelease.go
```diff
@@ -177,46 +195,70 @@ func createGitTag(ctx context.Context, client services.GitService, gitRepoInfo *
...
+		err := createTagForImage(ctx, *repo, quaySrcTag, *tag, quayTokenFilename, commitSHA)
 		if err != nil {
 			ok = false
 			fmt.Println("Failed to create the image tag for", r, "due to error:", err)
 		} else {
-			fmt.Println(fmt.Sprintf("Image tag '%s' created from tag '%s' with commit '%s' in repo '%s'", *tag, quaySrcTag, commitSHA, *repo))
+			fmt.Printf("Image tag '%s' created from tag '%s' with commit '%s' in repo '%s'\n", *tag, quaySrcTag, commitSHA, *repo)
 		}
 	}
+	os.Remove(f.Name())
```
laurafitzgerald, line 223: Unhandled Error here.

cmd/tagRelease.go
```diff
@@ -177,46 +195,70 @@ func createGitTag(ctx context.Context, client services.GitService, gitRepoInfo *
...
-	})
-	if err != nil {
-		return err
-	}
-	if len(tags.Tags) == 0 {
-		return fmt.Errorf("tag %s doesn't exit", quaySrcTag)
-	}
-	tag := tags.Tags[0]
-	commitID, _, err := quayClient.Manifests.ListLabels(ctx, quayRepo, *tag.ManifestDigest, &quay.ListManifestLabelsOptions{Filter: commitIDLabelFilter})
+func createTagForImage(ctx context.Context, quayRepo string, quaySrcTag string, quayDstTag string, quayTokenFilename string, commitSHA string) error {
```
laurafitzgerald, line 227: We don't use ctx anymore so it can be removed as a parameter

cmd/tagRelease.go
```diff
@@ -177,46 +195,70 @@ func createGitTag(ctx context.Context, client services.GitService, gitRepoInfo *
...
 
-func tryCreateQuayTag(ctx context.Context, quayClient *quay.Client, quayRepos string, quaySrcTag string, quayDstTag string, commitSHA string) bool {
+func tryCreateQuayTag(ctx context.Context, quayRepos string, quaySrcTag string, quayDstTag string, quayToken string, commitSHA string) bool {
 	repos := strings.Split(quayRepos, ",")
 	ok := true
+
+	//convert quayToken to a temp. config file.
+	f, err := os.CreateTemp("", "tmpfile-")
+	if err != nil {
+		fmt.Println("Failed to create quayToken temp file due to error:", err)
```
laurafitzgerald, line 205: If a failure here won't allow later commands to work as expected I think we can return false here.

laurafitzgerald: Or alternatively create another function where we do the quayToken conversion. The use of a bool return type here is because the function was solely a loop before where we didn't want to break out of the loop if one failure happened, but log it and continue and return a false if any one tag creation failed.

cmd/tagRelease.go
laurafitzgerald: IT can be remmoved up the chain of method calls as well.

cmd/tagRelease.go
```diff
@@ -177,46 +195,70 @@ func createGitTag(ctx context.Context, client services.GitService, gitRepoInfo *
...
-	if len(commitID.Labels) == 0 {
-		return fmt.Errorf("label %s doesn't exist", commitIDLabelFilter)
-	}
-	if *commitID.Labels[0].Value != commitSHA {
+	var imageInfo ImageInfo
+	json.Unmarshal(buf.Bytes(), &imageInfo)
+	commitID := imageInfo.Config.InnerConfig.Labels.CommitId
+	if commitID != commitSHA {
 		return fmt.Errorf("can't find an image with given tag %s that matches the given commit SHA: %s", quaySrcTag, commitSHA)
 	} else {
```
laurafitzgerald, line 247: The else could be removed here as the return breaks out anyway if the if is true.

cmd/tagRelease.go
laurafitzgerald, line 243: Unhandled error here

laurafitzgerald (changes requested):
Hi @martinhesko while looking at the unit tests I. noticed some possible improvements in the code so leaving this review. After some small updates the test is passing but I think we can improve them still. Taking a look at that now. :)

cmd/tagRelease.go
martinhesko: the `quayRepo` variable needed to be split for `mirror.Mapping` further down anyway and using it here saved me some times with testing it on my private quay (I could easily just replace `quayOrg` with `integreatly` when it's split into org and image) so I would probably keep it this way

cmd/tagRelease_test.go
```diff
@@ -514,7 +325,7 @@ func TestDoTagRelease(t *testing.T) {
 	for _, c := range cases {
 		t.Run(c.desc, func(t *testing.T) {
 			gitRepo := &githubRepoInfo{repo: DefaultIntegreatlyOperatorRepo, owner: DefaultIntegreatlyGithubOrg}
-			err := DoTagRelease(context.TODO(), c.ghClient, gitRepo, c.quayClient, c.tagReleaseOptions)
+			err := DoTagRelease(context.TODO(), c.ghClient, gitRepo, "cmFuZG9tdG9rZW4=", c.tagReleaseOptions)
```
austincunningham, line 328: this is some random string

austincunningham: @laurafitzgerald ?

laurafitzgerald: it's base64 encoding of a test value.

austincunningham: /lgtm

PR #266, opened 2022-06-08
MGDAPI-3970: Added support for custom SMTP parameters for addon-flow installations

martinhesko:
# Issue link
[MGDAPI-3970](https://issues.redhat.com/browse/MGDAPI-3970)

# What
Added the custom SMTP parameters as an option when installing through addon flow according to the specification [here](https://gitlab.cee.redhat.com/service/managed-tenants/-/merge_requests/2235/diffs). Also vim went crazy and removed a bunch of unneeded whitespace around the script :shrug: 

# Verification steps
Described in the main gitlab PR


trepel:
@martinhesko Firstly, the ci/prow/unit failed due to ` https://www.shellcheck.net/wiki/SC2236 -- Use -n instead of ! -z.`

Secondly (minor) I was slightly confused since we have `cluster/prepare/smtp` and `cluster/prepare/quota`. In the former the dummy redhat-rhoam-smtp secret is created. In the latter the `addon-managed-api-service-parameters` secret with quota and custom SMTP params is created. I can't help feeling we should rename at least the cluster/prepare/quota somehow.

cc @Boomatang in case you have some cool idea about the potential new name :)

trepel:
Anyway, I created a cluster and installed RHOAM via addon flow using this PR and it worked as expected so I think it is good to go once that unit test check is fixed (that renaming is optional I would say)

Boomatang:
@trepel on the renaming of the make target. Yes, I agree but also it would have a wider reach. All the pipelines would need to be checked, not sure if it would have an effect on CPaaS and from the developer point it could be affect work flows. As it is setting blank values for the custom smtp by default I still feel the name is more related to the quota work. I am 50/50 on needing to change the name.

trepel:
@martinhesko @Boomatang sorry guys, it seems I placed my comments randomly between delorean and integreatly-operator PR :facepalm: 

Again and better hopefully. For this one the code looks good to me. Just fix that prow/unit check and the suggestion from @Boomatang . The RHOAM installation using the PR was successful.

trepel:
As for renaming (that actually belongs to the https://github.com/integr8ly/integreatly-operator/pull/2719) I agree it's optional so you can just ignore it given how much additional work and testing it would require.

PR #275, opened 2022-07-22
MGDAPI-4369 - update oc version to 4.11

adam-cattermole:
[MGDAPI-4369](https://issues.redhat.com/browse/MGDAPI-4369)

This should be reverted to stable once 4.11 reaches this state.


adam-cattermole:
(force pushed update from rc4 -> rc5)

PR #280, opened 2022-08-22
MGDAPI-4479 fix commit message to contain MGDAPI-4480 reference

MStokluska:
Verify that the commit message contains correct jira reference to pass commits checks on intly repo.

https://issues.redhat.com/browse/MGDAPI-4480


KevFan:
Verified that the commit message container the `MGDAPI-4480` Jira reference :+1: 

/lgtm

PR #281, opened 2022-08-23
MGDAPI-1729 add delay between rosa role and sts cluster creation

cecobask:
The AWS roles created for ROSA to use seem to be a bit slow to apply and randomly cause pipelines to fail when running the `rosa create cluster` command with the following error:
```
ERR: Failed to retrieve AWS regions: Failed to assume role with ARN 'arn:aws:iam::***:role/ManagedOpenShift-Installer-Role' : operation error STS: AssumeRole, https response error StatusCode: 403, RequestID: ***, api error AccessDenied: User: arn:aws:sts::***:assumed-role/RH-Managed-OpenShift-Installer/OCM is not authorized to perform: sts:AssumeRole on resource: arn:aws:iam::***:role/ManagedOpenShift-Installer-Role
``` 
Adding a short pause between the commands fixed this issue.

### DIFF
--- a/scripts/sts/sts.sh
+++ b/scripts/sts/sts.sh
@@ -51,6 +51,7 @@ MACHINE_TYPE="${MACHINE_TYPE:-m5.xlarge}"
 provision_sts_cluster() {
     rosa login --env=$OCM_ENV
     rosa create account-roles --mode auto -y
+    sleep 30s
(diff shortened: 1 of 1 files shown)


KevFan:
Looks good to me :+1: 

/lgtm

PR #282, opened 2022-08-23
MGDAPI-4482 - add opm to delorean image

adam-cattermole:
[MGDAPI-4482](https://issues.redhat.com/browse/MGDAPI-4482)

`opm` is missing on the delorean container. The auto builds of delorean-cli container are using ubi as a base image, resulting in glibc 2.28 (required by opm) to be present.

Installing `opm` would work alone for the automatic builds, but  `opm` would not work on any local build from centos:7. I've opted for moving to the ubi image instead for local builds also.


KevFan:
Changes looks good to me and can build image successfully 👍 

/lgtm

PR #286, opened 2022-10-04
Update product version independently of operator

adam-cattermole:
Kept the interaction the same, but added a new optional parameter to update the product version independently of the operator version. This is required by EWS

### DIFF
--- a/cmd/manageRHMITypes.go
+++ b/cmd/manageRHMITypes.go
@@ -12,10 +12,16 @@ import (
 )
 
 type manageTypesCmdOptions struct {
-	filepath string
-	product  string
-	version  string
+	filepath        string
+	product         string
+	operatorVersion string
+	productVersion  string
 }
+
+const (
+	OperatorVersionType = "OperatorVersion"
+	ProductVersionType  = "Version"
+)
 
 func init() {
 
@@ -23,9 +29,9 @@ func init() {
 
 	cmd := &cobra.Command{
 		Use:   "set-product-operator-version",
-		Short: "Sets the operator version for a product in the rhmi_types file and sets product to CHANGEME if a minor version update",
+		Short: "Sets the operator version for a product in the rhmi_types file and sets product version to product-version if supplied",
 		Run: func(cmd *cobra.Command, args []string) {
-			err := SetVersion(flags.filepath, flags.product, flags.version)
+			err := SetVersion(flags.filepath, flags.product, flags.operatorVersion, flags.productVersion)
 			if err != nil {
 				handleError(err)
 			}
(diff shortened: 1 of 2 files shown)


cmd/manageRHMITypes.go
```diff
@@ -40,13 +46,15 @@ func init() {
 	cmd.Flags().StringVarP(&flags.product, "product", "p", "", "The product name")
 	cmd.MarkFlagRequired("product")
 
-	cmd.Flags().StringVarP(&flags.version, "version", "v", "", "The desired version")
+	cmd.Flags().StringVarP(&flags.operatorVersion, "version", "v", "", "The desired operator version")
 	cmd.MarkFlagRequired("version")
+
+	cmd.Flags().StringVarP(&flags.productVersion, "product-version", "e", "", "The desired product version")
```
cecobask, line 52: Wondering if we can use flags that are more meaningful here?
Something like `ov` instead of `v` for operator version and `pv` instead of `e` for product version.

cecobask (commented):
Verified that `delorean` cli works as expected in `integreatly-operator` with the provided command. One comment to address, other than that it looks good to me 👍🏻

cmd/manageRHMITypes.go
adam-cattermole: I was also hoping this was possible but unfortunately the library we use only supports single ascii character flags. Given that requirement I thought it would be easier not to change the existing flag so I don't need to change all uses of the function.

For ref:

```go
panic: "ov" shorthand is more than one ASCII character

goroutine 1 [running]:
github.com/spf13/pflag.(*FlagSet).AddFlag(0xc000506800, 0xc0002e7180)
        /home/adam/go/src/github.com/integr8ly/delorean/vendor/github.com/spf13/pflag/flag.go:864 +0x4ec
github.com/spf13/pflag.(*FlagSet).VarPF(0xc000506800?, {0x2afc558, 0xc0003bb520}, {0x2707f0e, 0x7}, {0x2702482, 0x2}, {0x2734f05, 0x1c})
        /home/adam/go/src/github.com/integr8ly/delorean/vendor/github.com/spf13/pflag/flag.go:831 +0x151
github.com/spf13/pflag.(*FlagSet).VarP(...)
        /home/adam/go/src/github.com/integr8ly/delorean/vendor/github.com/spf13/pflag/flag.go:837
github.com/spf13/pflag.(*FlagSet).StringVarP(0x7?, 0x27021e0?, {0x2707f0e?, 0xc0003bb4c0?}, {0x2702482?, 0xc000135400?}, {0x0?, 0x20dabe0?}, {0x2734f05, 0x1c})
        /home/adam/go/src/github.com/integr8ly/delorean/vendor/github.com/spf13/pflag/string.go:42 +0x6e
github.com/integr8ly/delorean/cmd.init.14()
        /home/adam/go/src/github.com/integr8ly/delorean/cmd/manageRHMITypes.go:49 +0x1f7
```

cecobask: Thank you for clarifying that for me!
I'm happy with the changes then 👍🏻 

/lgtm

PR #294, opened 2023-01-13
expose byovpc with private link option for ROSA clusters

laurafitzgerald:
JIRA

https://issues.redhat.com/browse/MGDAPI-5065

WHAT
Exposes the options for deploying ROSA clusters to 
BYOVPC and Private Link Clusters.
If Multi AZ is true then there must be compute notes divisible by 3
and subnets must be provided.
In the case of BYOVPC the machine cidr must match the cidr value of the VPC so we need to allow that to be configured as well. 
If Private Link is used then the region being used much match a set that is allowable in staging environment for OCM. There is a JIRA linked about that and also some information linked in the JIRA for this ticket.

### DIFF
--- a/scripts/rosa/rosa.sh
+++ b/scripts/rosa/rosa.sh
@@ -65,6 +65,10 @@ MIN_REPLICAS="${MIN_REPLICAS:-4}"
 MAX_REPLICAS="${MAX_REPLICAS:-6}"
 STS_ENABLED="${STS_ENABLED:-true}"
 MULTI_AZ="${MULTI_AZ:-false}"
+PRIVATE_LINK="${PRIVATE_LINK:-false}"
+BYOVPC="${BYOVPC:-true}"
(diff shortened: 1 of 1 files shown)


trepel:
I tried to create a cluster using this PR (ROSA byovpc multiAZ --private-link=true). The byovpc was created beforehand. Private subnets used for the ccommand:
`make ocm/rosa/cluster/create STS=true AWS_REGION=us-east-1 PRIVATE_LINK=true MULTI_AZ=true COMPUTE_NODES=6 SUBNET_IDS=subnet-07ddbee27c35114e8,subnet-0ef2f3782f6a1ca4e,subnet-0c51cb6294c412e78 CLUSTER_NAME=rspl MACHINE_CIDR=10.0.0.0/16 BYOVPC=true`

It went OK, I was also able to install RHOAM on the cluster.

/lgtm
/approve

PR #295, opened 2023-01-16
MGDAPI-4024 - updated dockerfile to handle generating manifests

martinhesko:


trepel:
I built the image and checked that git, diffutils, and ssh commands could be executed. I also checked that /.ssh dir was present.

/lgtm

PR #296, opened 2023-01-24
MGDAPI-5100 - fixed false positive for 'unhealthy' status

martinhesko:


trepel:
I just tested via echo "unhealthy" | grep -q -E ^healthy echo $? # was 1

echo "healthy" | grep -q -E ^healthy echo $? # was 0

ocm get subs --parameter search="cluster_id = ''" | jq -r '.items\[0\].metrics\[0\].health_state' | grep -q -E ^healthy echo $? # was 0 since I only have healthy clusters atm

/lgtm

PR #299, opened 2023-02-02
Added VPC template for ROSA BYOVPC private-link cluster

trepel:
Added VPC template for ROSA BYOVPC private-link cluster. This is required for instance for A45  test case, see https://github.com/integr8ly/integreatly-operator/pull/3095

Verification Steps

Eye review. Double check that no sensitive info is there.


martinhesko:
read through the template and there doesn't seem to be any sensitive info left,

/lgtm

PR #314, opened 2023-04-11
Added cidr-range param

trepel:
Added support for specifying the cidr-range param for RHOAM addon flow installations


trepel:
Validated via Playground here:
https://master-jenkins-csb-intly.apps.ocp-c1.prod.psi.redhat.com/job/Playground/job/trepel/job/rhoam-addon-flow-gcp/13/console

As you can see in the build log the CIDR was properly set to 10.1.0.0/22, maintenance-day to 1 and maintenance-hour to 0 and the addon installation triggered just fine

As you can see in the build log the

PR #319, opened 2023-05-31
MGDAPI-5766 - script to pull the data from a cluster

valerymo:
https://issues.redhat.com/browse/MGDAPI-5766
Script to pull the described data from a cluster

## Validation
Run scrip example:
```
cd delorean
$ scripts/delorean/get-cluster-data.sh
```
Output example:
**** Get cluster data - Tue Jun  6 11:32:50 IDT 2023 ****
* API calls being sent through RHOAM: 12014
* User CRs count: 1
* Backend count: 2
* Product count: 3
* Application plan count: 4
* Application count: 4
* Activedoc count: 4
* Policy created count: 0
* Authentication Providers Developer Portal count: 2
* Tenant count: 1
****  Please see logs in  ./cluster_data_log.txt ****

Script can be executed several times. Data/log will be added to ./cluster_data_log.txt file

Log file example:
```
$ cat ./cluster_data_log.txt
**** Get cluster data - Tue Jun  6 11:33:56 IDT 2023 ****
* API calls being sent through RHOAM: 12018
* User CRs count: 1
* Backend count: 2
* Product count: 3
* Application plan count: 4
* Application count: 4
* Activedoc count: 4
* Policy created count: 0
* Authentication Providers Developer Portal count: 2
* Tenant count: 1
****  Please see logs in  ./cluster_data_log.txt ****

**** Get cluster data - Tue Jun  6 11:35:01 IDT 2023 ****
* API calls being sent through RHOAM: 12022
* User CRs count: 1
* Backend count: 2
* Product count: 3
* Application plan count: 4
.....
....
```

### DIFF
--- a/scripts/delorean/get-cluster-data.sh
+++ b/scripts/delorean/get-cluster-data.sh
@@ -12,7 +12,12 @@
 # if they are using the developer portal
 # policy created count
 # Run script Example:  ./scripts/delorean/get-cluster-data.sh
-# Need be logged into the Openshift cluster before running this script
+# Need be logged into the Openshift cluster as kubeadmin before running this script
+
+# Please note that following counts provided only for admin tenant:
+# application_plans, services / products, backend_apis, active_docs, policies, authentication_providers/dev portal.
+# Tenants access token can't be retrieved by script;
+# provider_verification_key, that probably should be used instead of Tenant account token -  is not working, maybe 3scale bug.
 
 
 oc_whoami=$(oc whoami)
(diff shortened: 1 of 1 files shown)


scripts/delorean/get-cluster-data.sh
```diff
@@ -0,0 +1,98 @@
...
+oc_whoami=$(oc whoami)
+if [[ -z "${oc_whoami}" ]]; then
+  echo "Need be logged into the Openshift cluster before running this script. Please check: oc whoami. Exiting";
+  exit
+fi
+
+logfile="./cluster_data_log.txt"
+touch $logfile
+
+threescale_url=$(oc get route -n redhat-rhoam-3scale |grep admin |awk '{print $2}')
```
Boomatang, line 27: If there are multiply tenants this returns a list, which causes failures later in the script.

scripts/delorean/get-cluster-data.sh
```diff
@@ -0,0 +1,98 @@
...
+#echo "threescale_url: ${threescale_url}"
+threescale_access_token=$(oc get secret system-seed -oyaml -n redhat-rhoam-3scale |grep ADMIN_ACCESS_TOKEN |awk '{print $2}' |base64 -d)
```
Boomatang, line 29: This access token only return the data from the default tenant. If there is more than one tenant the backends and products will not be recorded.

scripts/delorean/get-cluster-data.sh
```diff
@@ -0,0 +1,98 @@
...
+echo "* Policy created count: ${count}" |tee -a $logfile
+
+# Developer Portal
+# 3scale API name: Authentication Providers Developer Portal List
+count=$(curl -sb  GET "https://${threescale_url}/admin/api/authentication_providers.json?access_token=${threescale_access_token}" |grep -o system_name |wc -l)
+echo "* Authentication Providers Developer Portal count: ${count}" |tee -a $logfile
+
+# tenant
+# Check Tenant CRs count in all namespaces
+count=$(oc get Tenants --all-namespaces |wc  -l)
```
Boomatang, line 89: Tenants could have being created through the UI before the declarctive CR work was introduced in 3scale. If the case there would be no Tenant CR's. All so the default tenant does not have a CR.

scripts/delorean/get-cluster-data.sh
```diff
@@ -0,0 +1,98 @@
...
+# Run script Example:  ./scripts/delorean/get-cluster-data.sh
+# Need be logged into the Openshift cluster before running this script
+
+
+oc_whoami=$(oc whoami)
+if [[ -z "${oc_whoami}" ]]; then
+  echo "Need be logged into the Openshift cluster before running this script. Please check: oc whoami. Exiting";
+  exit
+fi
+
```
Boomatang, line 23: Would be nice to log the user and server output from the oc whoami. It would help with linking stats to a cluster.

Boomatang (changes requested):
If there is more than on tenant the script does execute correctly due to the number of admin urls. The access token for the API will only work on the default tenant. This means the stats from other tenants is not found. 

Not all clusters will have tenants defined by CR, the script should allow for both. 

I have not checked the developer portal counts are working correctly. I was using an non CCS cluster.

scripts/delorean/get-cluster-data.sh
valerymo: Fixed. this url now - only for admin
Thank you

scripts/delorean/get-cluster-data.sh
valerymo: Thank you for comment. Changed. Tenants now calculated as accounts, using 3scale API

valerymo:
Closing this PR, as was decided to move the script into integreately_help. 
This is new request/MR: https://gitlab.cee.redhat.com/rhcloudservices/integreatly-help/-/merge_requests/904/diffs

PR #323, opened 2023-06-19
MGDAPI-5854 - go: v1.19

KevFan:
* Update to use go 1.19
  *  `vendor` prow test will fail until this PR gets merged and the prow configuration is auto updated to use the new go version


adam-cattermole:
Changes look good and vendor is correct locally with go1.19

/lgtm

PR #333, opened 2024-02-27
Python update in VPC stack

trepel:
Python update in VPC stack. The VPC stack creation fails since Python v3.7 is no longer supported in AWS. They suggest to use v3.12. Also added various URLs to ALLOWLIST as suggested in https://access.redhat.com/documentation/en-us/red_hat_openshift_service_on_aws/4/html/prepare_your_environment/rosa-sts-aws-prereqs#osd-aws-privatelink-firewall-prerequisites_rosa-sts-aws-prereqs

Verification Steps

Eye review. I have just tried to create a Stack out of the template including modifications from this PR (called `prlink`) and that worked well:
```
$ aws cloudformation describe-stacks --stack-name prlink --query Stacks[0].StackStatus --output text
CREATE_COMPLETE
```


carlkyrillos:
Reviewed changes to template
/lgtm

